PR #11, opened 2015-11-18
Make KuzzleDocument.save options parameter optional

benoitvidis:


scottinet:
I KNEW I was going to miss one, so I checked my code twice. Didn't help much, obviously. :-p

+1

PR #29, opened 2015-12-09
add bower install in doc + remove promise vs callback

AnthonySendra:
* Add bower install in doc
* Remove block about Callback vs Promise in doc
* Replace exemple in NodeJS with promise


xbill82:
:+1:

PR #34, opened 2015-12-18
 #kuz-339 - add support for multi-index

benoitvidis:


scottinet:
I wonder if we really should default this argument to ``mainindex`` instead of making it mandatory with no default value. 

On the top of my hat, I can see pros and cons for both of these approach, so... +1. :-)

PR #38, opened 2016-01-07
add login/logout method to sdk

stafyniaksacha:


dist/kuzzle.js
```diff
@@ -597,6 +622,52 @@ Kuzzle.prototype.connect = function () {
   return this;
 };
 
+
+Kuzzle.prototype.login = function (strategy, credentials, expiresIn) {
```
jenow: You forgot to use the strategy parameter

jenow:
Should add a callback to the 'login' method and should return 'this' in 'login' and 'logout'

scottinet:
+1 & merge

PR #54, opened 2016-02-08
Check the JWT Token validity upon reconnection

scottinet:
Before emitting a `reconnect` event, if there is a JWT Token, the SDK checks its validity using the new `checkToken` API route.
If the token is invalid, the `jwtToken` property is emptied and a `jwtTokenExpired` event is fired.


xbill82:
:+1: everything seems neat :)

stafyniaksacha:
+1

ballinette:
+1

Issue #53, opened 2016-02-07
Subscribe just after new Kuzzle


AnthonySendra:
Hello,

When I try to make a subscribe just after kuzzle creation, my callback is never called.

```js
var kuzzle = new Kuzzle('http://localhost:7512', {defaultIndex: 'mainindex'})
kuzzle
  .dataCollectionFactory('paint')
  .subscribe({}, function (error, result) {
    console.log(result) // never called
  });
```

And when I take the same code and add a delay between creation and subscribe, everything goes well

```js
var kuzzle = new Kuzzle('http://localhost:7512', {defaultIndex: 'mainindex'})
setTimeout(function () {
  kuzzle
    .dataCollectionFactory('paint')
    .subscribe({}, function (error, result) {
      console.log(result) // called
    });
}, 1000);
```

I've also tried to play with `autoQueue `and `autoReplay` but still doesn't work
And no error or warning output.

AnthonySendra:
Oh, and when it works, the `result` parameter in the callback is not a object `KuzzleDocument`. Is it normal ?

scottinet:
I'll take a look after my current task.

scottinet:
Ok found it.
This is a stupid bug really, caused by the "do not renew multiple times before soem delay has expired (approx. 200ms)" feature.

Here is what happens:

- The `KuzzleRoom` object stores the current timestamp 
- During subscription, the `KuzzleRoom` object checks the SDK status. As it is not yet connected, it registers itself to the main `Kuzzle` object and leaves.
- Once the SDK has connected, the `Kuzzle` object renew all pending subscriptions.
- We return to the `KuzzleRoom`. But because there wasn't 200ms since the last try, the object thinks it is too early to renew the susbcription and leaves.

I'm gonna deploy a quick fix in a few minutes, and I'll add some unit tests to check against this specific case.

scottinet:
As for your second question, receiving a plain-old JSON object as a notification is perfectly normal for the moment.
The `KuzzleRoom` object doesn't interpret notifications, as they can be of different types (document or users notifications).
Moreover, even if we resolve the result with a `KuzzleDocument` object in case of a document notification, the end-user would lose the `state` and `scope` informations.

However there @jenow and I discussed this issue some time ago, and we concluded that adding a `KuzzleNotification` object would be nice. There is a high probability that we do that after Kuzzle go beta.

scottinet:
Issue resolved. :-)

PR #55, opened 2016-02-08
Fixed the bug raised in the issue #53

scottinet:
- The `KuzzleRoom` object now stores the subscription timestamp when the subscription succeeds, instead of at the beginning of the `renew` function
- Updated unit tests to make sure the bug raised by the issue #53 doesn't happen again


xbill82:
:+1:

stafyniaksacha:
+1

benoitvidis:
+1

PR #56, opened 2016-02-08
made login and logout not queuable

scottinet:
* `Kuzzle.login()` and `Kuzzle.logout()` are now non-queuable, meaning they will never be queued during offline-mode
* When non-queuable queries are executed while been in offline-mode, if a callback is provided, then it is called with a "discarded query" error, instead of been silently discarded


xbill82:
:+1:

stafyniaksacha:
+1

benoitvidis:
+1

PR #57, opened 2016-02-08
Kuz406 refresh returns new object

scottinet:
* `KuzzleDocument.refresh` now resolves to a new instanciated `KuzzleDocument` object, instead of a refreshed version of itself
* `KuzzleDocument.refresh` now returns nothing, to stop call chains
* Removed the refreshing/queueing management from `KuzzleDocument`
* Updated unit tests


xbill82:
:+1:

benoitvidis:
+1

PR #58, opened 2016-02-09
Security object

AnthonySendra:
* CRUD User
* CRUD Profile
* CRUD Role


stafyniaksacha:
+1111

src/security/kuzzleUser.js
```diff
@@ -0,0 +1,134 @@
...
+ * @param {responseCallback} [cb] - Handles the query response
+ *
+ * @returns {*} this
+ */
+KuzzleUser.prototype.save = function (cb) {
+  var
+    data = this.serialize(),
+    self = this;
+
+  self.kuzzle.query(this.kuzzleSecurity.buildQueryArgs('createOrReplaceUser'), data, null, function (error) {
```
benoitvidis: nitpicking: For all calls to kuzzle.query, shouldn't we include the optional `options` in the query?

AnthonySendra: No you're right... But I think it's more clear to call the function with null instead. It works better with my IDE :)
But if you are convinced, I can change it. What do you think?

scottinet: To avoid any warning raised by your IDE, skipping the `options` argument altogether should do the trick: I made this argument optional in the JSDoc of the `query` method.

But another approach would be to add an optional `options` argument to this method (and maybe other methods too). Because by default, nearly all methods should support the `queuable` option, asking to (not) queue this query during offline mode.
The only exception to this rule I can think of should be the `login` and `logout` functions (and all other functions not sending any requests, like `addListener`, obviously).

src/kuzzle.js
```diff
@@ -231,7 +232,15 @@ module.exports = Kuzzle = function (url, options, cb) {
     }
   });
 
-  /*
+  /**
+   * Create an attribute security that embed all methods for manage Role, Profile and User
```
scottinet: Nitpicking: -for +to

scottinet:
Apart from the `queuable` option thing that we need to discuss: BIG +1

xbill82:
:+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

xbill82:
I'm the third validation, should I merge or would you like to discuss about the `queuable` option?

scottinet:
@AnthonySendra and I discussed the matter, and he agreed to add the `queuable` option to (almost) all these new methods, to be on par with the rest of the SDK.

New commits should be soon submitted

AnthonySendra:
@scottinet could you have a new look at this PR please ?

scottinet:
lgtm!

PR #60, opened 2016-02-12
fix hydrate user and replace in setContent

AnthonySendra:
* Fix hydrate user
* Fix hydrate profile
* Add option `replace` in setContent


stafyniaksacha:
+1

benoitvidis:
+1

PR #64, opened 2016-02-23
checkToken does not take options

scottinet:
The `Kuzzle.checkToken` method should not take options, and it should always be non-queuable.
Moreover, instead of returning the raw API response, it now returns only the informations concerning the token check.

The SDK documentation will be updated accordingly.


benoitvidis:
+1

stafyniaksacha:
+1

PR #65, opened 2016-02-24
Missing properties

scottinet:
We exposed the raw JSON content of the `KuzzleProfile` and `KuzzleUser` objects, but there were no obvious way to get the associated roles/profiles of these objects. 
As this could be confusing for SDK users, and because in strongly-typed languages associated roles/profiles are kept away from the raw content, I added two new specific getters.

The SDK documentation will be updated accordingly.


ballinette:
+1

dbengsch:
+1

benoitvidis:
+1

PR #67, opened 2016-03-01
Login credentials now optional

jenow:


src/kuzzle.js
```diff
@@ -434,6 +434,11 @@ Kuzzle.prototype.login = function (strategy, credentials, expiresIn, cb) {
       strategy: strategy
     };
 
+  // don't need credentials in case of an OAUTH strategy
+  if (typeof credentials === 'number' || typeof credentials === 'string') {
```
scottinet: Does not work if this method is called with only a strategy and a callback.
I suggest prototyping this function only with the `strategy` argument, and then going through the `arguments` global variable to assign other arguments depending on their type.

test/kuzzle/constructor.test.js
```diff
@@ -750,6 +750,26 @@ describe('Kuzzle constructor', () => {
         });
       });
 
+      it('should get the url redirection from OAUTH', function (done) {
```
scottinet: Since we're going to add some code to handle arguments, I suggest adding another unit test checking for all argument combinations.

PR #70, opened 2016-03-18
implemented offline queue loader and new events

scottinet:
Allow SDK developers to act on offline queue activity, the main purpose being to allow them to locally persist the queue if needed.

* fires a new `offlineQueuePush` event everytime a new request is queued (if it passes optional filters). The event  provides the pushed request and its associated callback to the event handler

* fires a new `offlineQueuePop` event everytime a request is deleted from the queue. The event provides the poped request to the event handler

* added a new SDK global property, called `offlineQueueLoader`. This property can be set with a function. If set, the provided function is called automatically to fill the offline queue before replaying it. Duplicates are removed, using the request ID as an unique key. Dequeuing can now throw exceptions if the property is improperly set or if the returned values are incorrect


benoitvidis:
+1

jenow:
+1

stafyniaksacha:
+1

PR #72, opened 2016-04-01
fixed setContent and update in KuzzleSecurityDocument

scottinet:
* The `KuzzleSecurityDocument.setContent` method was taking a `replace` argument, which is wrong: because of the way security document objects are constructed, this argument had no effect
* The `KuzzleSecurityDocument.update` method was updating the internal object content with the `content` argument instead of the updated source of the document


AnthonySendra:
+1

dbengsch:
+1

xbill82:
Seems Legit! +1

scottinet:
@xbill82 > well in fact, the SDK documentation didn't change in that regard. I implemented the `setContent` method in security objects by copying/pasting the one in `KuzzleDocument` without giving it a second thought, and that was clearly wrong.

So you may view this PR as a way to fix the SDK regarding the documentation. :-)

PR #74, opened 2016-04-06
hotfix: error event does not forward info

scottinet:
The `error` event wasn't forwarding anything to its listeners. This has been fixed: listeners now receive the connection error object.


xbill82:
+1

Issue #81, opened 2016-05-09
Refactor dataCollectionFactory signature


scottinet:
The arguments provided to the `dataCollectionFactory` method should be inverted, so that the optional `index` argument appears _after_ the required `collection` one

PR #86, opened 2016-05-24
fixes issue #81

scottinet:
Inverted the `index` and `collection` arguments of the `Kuzzle.dataCollectionFactory` method, to make the optional argument appear after the required one.


jenow:
+1

PR #88, opened 2016-05-27
KUZ-574 getRights methods to KuzzleSecurity

xbill82:
Added the `getMyRights` and `getUserRights` to the `KuzzleSecurity` object.

Question: shall we create a `userRights` object?


src/security/kuzzleSecurity.js
```diff
@@ -709,5 +709,115 @@ KuzzleSecurity.prototype.userFactory = function(id, content) {
...
+ * @param  {function} cb   The callback containing the normalized array of rights.
+ */
+KuzzleSecurity.prototype.getUserRights = function (userId, options, cb) {
+  var data = {_id: userId};
+
+  if (!userId || typeof userId !== 'string') {
+    throw new Error('userId parameter is mandatory for isActionAllowed function');
+  }
+
+  if (!cb && typeof options === 'function') {
```
jenow: Callback should be mandatory, otherwise there is no way to get the result.

src/security/kuzzleSecurity.js
```diff
@@ -709,5 +709,115 @@ KuzzleSecurity.prototype.userFactory = function(id, content) {
...
+  }
+};
+
+/**
+ * Gets the rights array of the currently logged user.
+ *
+ * @param  {function} cb The callback containing the normalized array of rights.
+ */
+KuzzleSecurity.prototype.getMyRights = function (options, cb) {
+  if (!cb && typeof options === 'function') {
```
jenow: Same here, callback should be mandatory, otherwise there is no way to get the result.

scottinet:
+1

src/security/kuzzleSecurity.js
xbill82: Done

stafyniaksacha:
+1

PR #92, opened 2016-06-14
[release-1.9.1] Fix #91: format correctly roles within profile definition

### LINKED ISSUE #91: Role refactor: fix KuzzleProfile format
After changes introduced into kuzzle by role refactoring (cf. https://github.com/kuzzleio/kuzzle/pull/232), we need to change the format of KuzzleProfile object in the SDK.

Profile definition are unitl now in following format: 
```
var profileDefinition = {
  roles: [
    'myrole', 'default'
  ]
};
```

We should change it to following format: 
```
var profileDefinition = {
  roles: [
    {_id: 'myrole' },
    {_id: 'default', restrictedTo: [ {index: 'index1’, collections: ['foo’, 'bar’]}, {index: 'index2’} ] }
  ]
};
```

ballinette:


benoitvidis:
+1

PR #96, opened 2016-07-04
fix #95: handle empty mappings

### LINKED ISSUE #95: KuzzleDataMapping.mapping is undefined when retrieving an empty mapping
It's possible to have empty mappings in elasticsearch:

```json
{
  "index": {
    "mappings": {
      "collection": {}
    }
  }
}
```

The SDK expects a `properties` attributes in the object returned by Elasticsearch, but in this case, there is none. And the SDK ends up building a `KuzzleDataMapping` object with an undefined mapping, causing unpredictable behaviors.

scottinet:
Fixes #95 + added unit tests to catch this kind of bug


benoitvidis:
:+1:

dbengsch:
+1

PR #100, opened 2016-08-02
WebSocket support

scottinet:
Add pure WebSocket capabilities to the Javascript SDK, while still supporting SocketIO to ensure compatibility with older browsers.

- isolated network calls within wrappers
- first tries to instantiate websocket objects and falls back to socket.io if unable to do so

## Unrelated changes

- requests history is now cleaned once every 1s, instead of once per request sent. Dramatically improves performances.
- Solve a crash when `Kuzzle.query` is invoked without a query argument. Now throws an error when such a case occurs
- Fixed `responseCallback` JSDoc


dbengsch:
+1

stafyniaksacha:
+1

dist/kuzzle.js
```diff
@@ -272,167 +393,47 @@
...
+ * @param {Object} [data] - The content of the query response
  */
 
 /**
  * Kuzzle object constructor.
- * @param url - URL to the Kuzzle instance
- * @param index - Database index
+ *
+ * @constructor
+ * @param address - Server name or IP Address to the Kuzzle instance
```
benoitvidis: Shouldn't we rather use `host` instead of `address`? 
Looks to me to be more standard and more accurate.

scottinet: Done. I've also fixed a potential bug with the `off` function in websocket wrappers (node + browser)

benoitvidis:
Apart from the host/address naming thing, +1

PR #102, opened 2016-08-10
Webpack packaging

xbill82:
This PR drops Grunt and sets Webpack as default package bundler for browser use.
Builds are launched via

    npm run build

and builds are dropped into `dist/`.

Built files still need some testing IMHO, since some external libraries are excluded (`noParse: [/wsnode\.js/, /crypto/, /buffer/, /node-uuid/]`) and I'm not sure everything works fine.


package.json
```diff
@@ -27,22 +28,29 @@
     "bluebird": "3.4.1",
     "bufferutil": "^1.2.1",
     "node-uuid": "1.4.7",
+    "ora": "^0.3.0",
     "utf-8-validate": "^1.2.1",
+    "webpack": "^1.13.1",
```
dbengsch: shouldn't webpack be a devDep ?

stafyniaksacha:
LGTM

package.json
xbill82: YEP

dbengsch:
+1

jenow:
+1

PR #108, opened 2016-09-06
KUZ-410: normalize chaining

scottinet:
* `.eslintrc` is now deprecated, using `.eslintrc.json` instead
* Removed chaining capabilities from functions with required callbacks
* Added chaining capabilities to functions with optional callbacks
* Simplified some callback handling, shaving some space from the minified library


src/kuzzle.js
```diff
@@ -588,13 +575,12 @@ Kuzzle.prototype.whoAmI = function (callback) {
...
     callback(null, new KuzzleUser(self.security, response.result._id, response.result._source));
   });
-
-  return self;
 };
 
 /**
  * Gets the rights array of the currently logged user.
  *
+ * @param {object} [options] - Optional parameters
```
ballinette: so the `options` parameter should be removed from the function definition?

scottinet: Fixed.

ballinette: :+1:

ballinette:
Need to clean-up the jsdoc:
remove `@returns {object} this` when the method no longer returns the chaining object.
Add it when the chaining has been added.

Despite of that, :+1: for me ;)

scottinet:
Dang, you're right about jsdoc :-(

It has been updated.

PR #111, opened 2016-09-08
Fix #110: Handle WebSocket close events

### LINKED ISSUE #110: Does not retry connecting through websocket on a connection error
The current implementation of the "onclose" event on a WebSocket connection does not check the [WebSocket close event code](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent).

This means that if a connection closes with any other code than 1000, the SDK still consider this event as a normal disconnection and does not retry connecting automatically.

scottinet:
Now checks the connection close event code and calls the onError listeners instead of the onDisconnect ones when appropriate.


src/networkWrapper/wrappers/wsbrowsers.js
```diff
@@ -234,4 +232,25 @@ function poke (listeners, roomId, payload) {
...
 }
 
+/**
+ * Called when the connection closes with an error state
+ *
+ * @param {boolean} autoReconnect
+ * @param {number} reconnectionDelay
+ * @param {string|Object} message
+ */
+function onClientError(autoReconnect, reconnectionDelay, message) {
```
ballinette: isn't it possible to factorize a little bit the code btw. wsbrowsers.js and wsnode.js ?
(as far as I can see, only the code of `this.connect()` method has some differences in its syntax)

ballinette:
+1

src/networkWrapper/wrappers/wsbrowsers.js
scottinet: It should be, I separated the two wrappers because I initially thought that there would be more differences than that.

Right now the biggest problem is with webpack, when it tries to package `ws` and other modules that are incompatible with web browsers (crypto and bufferutil).
The only solution I found was to exclude wsnode.js from webpack's analysis. We need to find some way to factorize wsnode and wsbrowsers while preventing webpack to include these modules.

PR #112, opened 2016-09-15
KUZ-672 subscribe error handler

scottinet:
Currently, if a subscription fails (for instance because incorrect filters have been provided), there is no way for the user to know about it.
Worse: there is an exception thrown, but since it is thrown asynchronously, there is no way it can be catched by the client.

The following is a change proposal of a better way to handle asynchronous errors for subscriptions:

1. Change the KuzzleRoom.renew prototype to accept a second optional callback (urgh)
    This callback is to be called to propagate the subscription result

2. Change the KuzzleDataCollection.subscribe and KuzzleDocument.subscribe methods: instead of returning a KuzzleRoom object, they should return this kind of object:

```js
{
  "onDone": function (callback) { ... }
}
```

3. The "onDone" function takes a callback and returns itself, to allow chaining. When a subscription ends, it invokes the provided callback with the KuzzleRoom object as an argument


src/kuzzleRoom.js
```diff
@@ -182,20 +184,20 @@ KuzzleRoom.prototype.renew = function (filters, cb) {
...
   if (self.kuzzle.state !== 'connected') {
-    self.callback = cb;
+    self.callback = notificationCB;
     self.kuzzle.subscriptions.pending[self.id] = self;
     return;
   }
 
   if (self.subscribing) {
-    self.queue.push({action: 'renew', args: [filters, cb]});
+    self.queue.push({action: 'renew', args: [filters, notificationCB]});
```
scottinet: :+1:

PR #115, opened 2016-09-20
[MASTER] KUZ-683 allow changing host

scottinet:
Instead of being read-only, the `host`, `ioPort` and `wsPort` attributes from the `Kuzzle` object should be writable to allow changing them between connections.
And when invoking `Kuzzle.connect`, if a network connection has previously been established, then `Kuzzle.disconnect` is called beforehand.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-sdk",
-  "version": "2.1.0",
+  "version": "2.1.1",
   "description": "Official Javascript SDK for Kuzzle",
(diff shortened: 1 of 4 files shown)


scottinet:
This PR merges the changes into the `master` branch, to allow the BO team to implement multi-environments support.

dbengsch:
Shouldn't ioPort and wsPort be writable as well ?

scottinet:
@dbengsch > saw that too when updating the Android SDK :p

PR #117, opened 2016-09-21
add queryError event

stafyniaksacha:


src/kuzzle.js
```diff
@@ -57,7 +57,8 @@ module.exports = Kuzzle = function (host, options, cb) {
         jwtTokenExpired: {lastEmitted: null, listeners: []},
         loginAttempt: {lastEmitted: null, listeners: []},
         offlineQueuePush: {listeners: []},
-        offlineQueuePop: {listeners: []}
+        offlineQueuePop: {listeners: []},
+        queryError: {lastEmitted: null, listeners: []}
```
scottinet: Adding the `lastEmitted` property automatically prevents this event being fired more than once every `this.eventTimeout` ms (currently set to 200ms).

We may not want to throttle this.

stafyniaksacha: Done

scottinet:
Do you plan to add this one to android and ios?

j33f:
I need to have the actual pending PRs on iOS SDK to be merged then I'll figure out how to implement it

stafyniaksacha:
It can be delayed, JIRA Ticket was written

PR #118, opened 2016-09-21
add unsetJwtToken() method to kuzzle object

stafyniaksacha:


scottinet:
Do you plan to add this one to android and ios?

PR #124, opened 2016-10-03
Do not logout on kuzzle.disconnect

AnthonySendra:


ballinette:
(NB: I guess we need also to report this fix to other SDKs ?)

src/kuzzle.js
```diff
@@ -1032,8 +1032,6 @@ Kuzzle.prototype.listIndexes = function (options, cb) {
 Kuzzle.prototype.disconnect = function () {
   var collection;
 
-  this.logout();
```
xbill82: Didn't you want to put a condition here based on an argument?

AnthonySendra: After a quick discussion with @scottinet and @jenow, we decided that if the user want to logout, he can directly call `kuzzle.logout()` explicitely.

PR #134, opened 2016-11-07
Implement scroll api

### LINKED ISSUE #130: Rename advancedSearch -> search

stafyniaksacha:
fixes #130

* Migrate `advancedSearch` method to `search`
* Implement `scroll` (see kuzzleio/kuzzle#516)
* Implement `createIndex`


src/kuzzleSearchResult.js
```diff
@@ -0,0 +1,170 @@
...
+ * @param {object} [searchArgs]
+ * @param {KuzzleSearchResult} [previous]
+ * @constructor
+ */
+function KuzzleSearchResult (dataCollection, total, documents, aggregations, searchArgs, previous) {
+  Object.defineProperties(this, {
+    // read-only properties
+    dataCollection: {
+      value: dataCollection,
+      enumerable: true
```
scottinet: should not be enumerable

src/kuzzleSearchResult.js
```diff
@@ -0,0 +1,170 @@
...
+    },
+    searchArgs: {
+      value: searchArgs || {},
+      enumerable: true
+    },
+    // writable properties
+    _previous: {
+      value: previous || null,
+      enumerable: true,
```
scottinet: should not be enumerable

src/kuzzleSearchResult.js
```diff
@@ -0,0 +1,170 @@
...
+      writable: true
+    },
+    _next: {
+      value: null,
+      enumerable: true,
```
scottinet: should not be enumerable

src/kuzzleSearchResult.js
```diff
@@ -0,0 +1,170 @@
...
+
+  return this;
+}
+
+
+/**
+ * @param cb
+ * @returns {*}
+ */
+KuzzleSearchResult.prototype.previous = function (cb) {
```
scottinet: As discussed together, I think this method should have only 1 defined behavior. To keep consistency, it may be better to make it take a callback, as it makes more sense for someone using our SDK without knowledge of the underlying mechanisms.

src/kuzzleSearchResult.js
```diff
@@ -0,0 +1,170 @@
...
+        options = Object.assign({}, this.searchArgs.options),
+        fetchedDocuments = this.documents.length,
+        previous = this;
+
+      if (this.searchArgs.filters.scroll) {
+        options.scroll = this.searchArgs.filters.scroll;
+      }
+
+      // check if we need to scroll again to fetch all matching documents
+      while (previous = previous.previous()) {
```
scottinet: As discussed together: it's way better to propagate the number of fetched documents throughout the construction of the chained list, starting from 0.
That way, we get it in O(1) instead of rebuilding it each time in O(n²)

src/kuzzleDataCollection.js
```diff
@@ -79,52 +80,16 @@ function KuzzleDataCollection(kuzzle, collection, index) {
...
-      documents: documents
-    };
-
-    if (result.result.aggregations) {
-      response.aggregations = result.result.aggregations;
-    }
-
-    cb(null, response);
-  });
+  console.warn('Usage of KuzzleDataCollection.advancedSearch is deprecated. Use KuzzleDataCollection.search instead');
```
dbengsch: The log would be triggered at each call of `advancedSearch` ? It might make a lot

src/kuzzleDataCollection.js
```diff
@@ -447,6 +445,128 @@ KuzzleDataCollection.prototype.replaceDocument = function (documentId, content,
...
+      result.result.total,
+      documents,
+      result.result.aggregations ? result.result.aggregations : [],
+      {options: options, filters: filters}
+    ));
+  });
+};
+
+/**
+ * Scroll into a search result
```
xbill82: Can we write a more detailed description for this method? (would be handy not to have to go to the docs)

scottinet: done

PR #136, opened 2016-11-07
Re-enable eslint & add support for test folder

dbengsch:


test/kuzzle/listenersManagement.test.js
```diff
@@ -28,46 +28,50 @@ describe('Listeners management', () => {
...
-      catch (e) { }
+      catch (e) {
+        // done
+      }
     });
 
     it('should throw an error when providing a non-function listener argument', function () {
       try {
         kuzzle.addListener('connected', 'bar');
         should.fail('success', 'failure', 'Should have failed to add a string listener', '');
```
benoitvidis: I am probably missing something but I don't see how this can test anything at all (not due to this pr, the code was there before).

`should.fail` just throws an assertion error and does not depend on the result of the previous command.

There are many of these around.

dbengsch: This code will not go into should.fail if the previous code is throwing an error (and therefor will succeed). It is actually testing that the code is throwing "properly". I agree to you it is a bit dirty to do it that way... I just needed to have non empty blocks for eslint *sig*

scottinet: I think I wrote that code. I had discovered `should.fail` existence and I wanted to play with it, instead of adding a callback to `it()` and invoke it with an error if the expected throw did not happen.

benoitvidis: I agree this PR is not the place to fix these should.fail cases but still, these tests do not test anything at all.

`should.fail` throws an assertion error and there is no restriction set on the type of the cought error. Wathever happens, we end up in this empty catch block.

Opened https://github.com/kuzzleio/sdk-javascript/issues/138 to tackle it.

PR #152, opened 2016-12-09
Adapt the SDK to the controller refactor

dbengsch:
linked to kuzzleio/kuzzle#567


README.md
```diff
@@ -4,7 +4,7 @@
 Official Kuzzle Javascript SDK
 ======
 
-This SDK version is compatible with Kuzzle 1.0.0-RC6 and higher
+This SDK version is compatible with Kuzzle 1.0.0-RC8 and higher
```
scottinet: As @benoitvidis said to me (and I agree): is this really necessary?
Since we won't have any real users until v1 is out?

README.md
```diff
@@ -140,6 +140,10 @@ kuzzle
 });
 ```
 
+## Building manually
+
+Clone this github repository and run ``npm run build``. A ``dist`` directory will be created, containing a plain browserified version of this SDK, and a minified version.
```
scottinet: I removed this section because I've rewritten the README. There is no "browserified" version anymore, and since this step is required to use the SDK in a browser, I've already detailed the procedure in the "Installation/Javascript" section.

I think this section should go.

dbengsch:
@scottinet Done

PR #172, opened 2017-01-18
Fix scroll action and next on searchResult

dbengsch:


src/Collection.js
```diff
@@ -511,6 +511,10 @@ Collection.prototype.scroll = function (scrollId, options, filters, cb) {
     options = {};
   }
 
+  if (!options.scroll) {
+    throw new Error('Collection.scroll: scroll is required');
```
stafyniaksacha: `scroll` parameter is not required to do scroll request (that is only to force refreshing scroll id)

ballinette: Hmmm are your sure ? the ES documentation gives always a `scroll` parameter in its examples:
https://www.elastic.co/guide/en/elasticsearch/reference/5.0/search-request-scroll.html
```
POST  /_search/scroll 
{
    "scroll" : "1m", 
    "scroll_id" : "DXF1ZXJ5QW5kRmV0Y2gBAAAAAAAAAD4WYm9laVYtZndUQlNsdDcwakFMNjU1QQ==" 
}
```

stafyniaksacha: As discussed together, LGTM

PR #177, opened 2017-02-08
Remove previous and next cache in SearchResult

### LINKED ISSUE #176: In SearchResult, previous and next features induce a serious memory leak

dbengsch:
Fix  #176

* Remove `_previous` and `_next` cache in `SearchResult`
* Remove `previous` function that could't be reworked properly for a scroll


src/SearchResult.js
```diff
@@ -1,11 +1,12 @@
 /**
- *
  * @param {Collection} dataCollection
  * @param {int} total
  * @param {Document[]} documents
  * @param {object} [aggregations]
  * @param {object} [searchArgs]
  * @param {KuzzleSearchResult} [previous]
+ * @property {Collection} dataCollection
```
ballinette: ???

dbengsch: This states the object built the constructor function has a defined property called dataCollection of type Collection

dbengsch:
@xbill82: 
Issue about the documentation created https://github.com/kuzzleio/documentation/issues/88 .
If you find something else that is not documented, please feel free to create an issue in the documentation repository (https://github.com/kuzzleio/documentation).

dbengsch:
@xbill82 Please refer to the connected ticket to know why we had to remove them.

xbill82:
@dbengsch Thanks, I actually didn't really read the title of the issue (just got annoyed by the "no description" message, but the title is actually enough :) I understand that the problem is that they introduce a memory leak, but I've missed all the discussion that lead to the decision to take them out instead of fixing the mem leak. We can discuss it IRL if u prefer.

src/SearchResult.js
ballinette: OK, my bad. I read "@param" instead of "@property"

benoitvidis:
@xbill82 This issue is severe enough imo to take immediate actions and remove this pseudo-functionality just right now.

Users can very easily work around the removed previous options on client side by storing the whole dataset in a local collection but this is not what Kuzzle should do on its side by default.

scottinet:
@xbill82 > this is not a memleak, since the used memory is not lost, but memory consumption. And that's the problem: there is no way to implement this feature without the client consuming too much memory.

It has been decided to make this cursor "forward-only", as are many cursor implementations out there.

xbill82:
@scottinet thanks for the details, now I get the point (and the severity) of the issue.

Issue #181, opened 2017-02-14
Rename the updateIfExist option


scottinet:
Currently, the `Collection.createDocument` method allows the `updateIfExist` option. Which does a document replacement if it exists in Kuzzle's database.

Since this option is misleading, and because we may want to add a `createOrUpdate` API route later, we propose the following:

* remove the `updateIfExist` option
* add a `ifExist` option, taking the following values: `error` (default), `replace`

PR #182, opened 2017-02-14
Collection.createDocument: rename the updateIfExist option

scottinet:
# Description

The `Collection.createDocument` method accepts a `updateIfExist` option.
Since this option really replaces the document if it already exists, it is dangerously confusing.

This PR change this option name to `ifExist` and its accepted value to one of these 2 values: `error` (default) or `replace`
If an invalid value is provided, the method throws.

# Solved issue

#181


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-sdk",
-  "version": "3.2.0",
+  "version": "3.3.0",
```
stafyniaksacha: Bad bump, should not be on this branch

scottinet: Fixed

PR #184, opened 2017-02-16
Browsers compatibility fix

scottinet:
# Description

While scanning this SDK code to fix an unrelated documentation issue, I came across a `process.nextTick` call, which is a pure-NodeJS method.
So, I fixed that. Which led to a failed unit test, also relying on `process.nextTick`.

It came to me that having a browser-compatible code tested with ES6 and/or NodeJS specific code was not a good idea, so I modified ESLint to make sure our code and tests are using the same standards, and reverted unit tests from ES6 to ES5.


xbill82:
I don't understand the point of this PR, since the code is transiped by Webpack and in the README we tell developers to use the transipled version in `dist` for the browser so... All the `nextTick` are automatically transpiled to `setTimeout(..., 0)` by Webpack, making the transpiled code ES5 compliant.

scottinet:
We certainly use webpack, but there is no transpilation, everything is in pure ES5, and webpack uses the code as is.

This was a choice made a long time ago, but now this is a choice I regret. One day I'll rewrite everything in ES6 (or even ES7 if I make my mind late enough) and use a transpiler.

xbill82:
Ok, I understand.
As an aside, I can reasonably say that Webpack can translate node-specific methods to browser-specific ones. This was the aim of Browserify, which Webpack reproduces.

PR #185, opened 2017-02-17
Send event discarded when no room provided

AnthonySendra:


src/networkWrapper/wrappers/websocket.js
```diff
@@ -72,6 +72,9 @@ function WSNode(host, port, ssl) {
       if (data.room && self.listeners[data.room]) {
         poke(self.listeners, data.room, data);
       }
+      else {
+        poke(self.listeners, 'discarded', data);
```
scottinet: Since it's private, the `poke` function considers it's the responsability of the caller to make sure the provided parameters are correct.
And `poke` will use the `self.listeners[room]` directly without verifying if it's defined or not, this you're supposed to call it only to poke existing listeners.

Long story short: this line of code will end up in a `TypeError` exception if there is no listener on the `discarded` event.

PR #188, opened 2017-02-27
Make all options given to kuzzle constructor writable

benoitvidis:
# Description

In some circumstances, we may want to use the same kuzzle object and change its options, for instance to switch to another environment.

This PR allows to update any possible option given to Kuzzle constructor.

Notably, this allows to switch the connection mode back from or to ssl on the current kuzzle instance, without having to create a new object.

# Linked issue

https://github.com/kuzzleio/kuzzle-backoffice/issues/154

# Linked PRs

https://github.com/kuzzleio/documentation/pull/113

### DIFF
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -95,9 +95,10 @@ function Kuzzle (host, options, cb) {
       },
       writable: true
     },
-    // read-only properties
+    // configuration properties
     autoReconnect: {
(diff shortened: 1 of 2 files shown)


scottinet:
This needs to be documented, with a visible warning explaining to users that changing these parameters won't take effect until an entirely new connection is created.

PR #194, opened 2017-03-10
Stop calling onClientError when disconnect is called

jenow:
Fix https://github.com/kuzzleio/sdk/issues/16


ballinette:
Indedd, the related issue will be fixed by another way when the network wrappers will be refactored.
Is the bug is blocking, we can accept this quick fix.
If not, it's better to wait for the deep network wapapers refactoring.

scottinet:
Ok, accepting this fix for now since it'll be implemented in a cleaner way later.

PR #197, opened 2017-03-16
Harmonize event emitters implementation

ballinette:
main `Kuzzle` SDK prototype and websocket wrapper implementation both used their own event listeners.

Now, both Kuzzle and websocket wrapper inherits from Node build-in `EventEmitter` class (https://nodejs.org/api/events.html), with a polyfilled implementation for browser.

## Breaking Changes
method `addListener` no longer returns an ID, to be compliant with NodeJS implementation ( https://nodejs.org/api/events.html#events_emitter_on_eventname_listener )
It now returns the Emitter object itself, so that it can be chained.

For each event, the listener callback function is used as the key in listeners list

## Documentation
https://github.com/kuzzleio/documentation/pull/172


src/Kuzzle.js
```diff
@@ -210,6 +197,8 @@ function Kuzzle (host, options, cb) {
     }
   }
 
+  this.eventEmitter = new EventEmitter(this.eventTimeout);
```
benoitvidis: Do we need a property to handle events? Wouldn't it make more sense if the main object (`Kuzzle` || `WSNode`) would be an event emitter itself?

src/eventEmitter/index.js
```diff
@@ -0,0 +1,165 @@
...
+    self = this,
+    now = Date.now(),
+    args = Array.prototype.slice.call(arguments, 1),
+    eventProperties = this.eventListeners[event];
+
+  if (!eventProperties) {
+    return false;
+  }
+
+  if (eventProperties.lastEmitted && eventProperties.lastEmitted >= now - this.eventTimeout) {
```
benoitvidis: This is one of the specific implementations for Kuzzle but for the main part, the rest of the implementation looks rather standard to me.

[discussion] Would it make sense to rely on an external library (i.e. webpack uses [ev-emitter](https://github.com/metafizzy/ev-emitter/blob/master/ev-emitter.js) by default), and override its functionalities when required?

ballinette: The code here is quite light, and anyway, we would have to override all library's methods to manage the `eventTimeout` behaviour.
So the question is: do we still need this behaviour or not ? 
If yes, I rather reimplement the whole methods like here
If not, of course, we can simply remove it and use `ev-emitter` instead.

benoitvidis: OK. I was just looking for more arguments in favor of relying on webpack more for the sdk, but this is probably not the best example.

benoitvidis: After discussion, one option could be nice: to use a library that mimics node `events` methods (`on`, `off`, `emit` etc). 
When running on node.js, the sdk could just use the native module and use the polyfill when targeting the browser.

src/Kuzzle.js
```diff
@@ -1,5 +1,6 @@
 var
   uuid = require('uuid'),
+  EventEmitter = require('./eventEmitter'),
```
stafyniaksacha: This is totally bad, we should not override build-in objects (Defined in nodejs context or browser one)

ballinette: The PR was still under progress.

You can review it now with the latest updates ;)

src/Kuzzle.js
```diff
@@ -48,19 +50,20 @@ function Kuzzle (host, options, cb) {
...
-        loginAttempt: {lastEmitted: null, listeners: []},
-        offlineQueuePush: {listeners: []},
-        offlineQueuePop: {listeners: []},
-        queryError: {listeners: []},
-        discarded: {listeners: []}
-      }
+    eventActions: {
+      value: [
+        'connected',
+        'kerror',
```
ballinette: `error` events have a special behaviour in Node, and throw an error if it is triggered with no listener (see https://nodejs.org/api/events.html#events_error_events )
So we need another event name for Kuzzle errors to avoid bad behaviour.

Looking for a better name, any suggestion is welcome ;)

src/Kuzzle.js
```diff
@@ -353,7 +361,7 @@ Kuzzle.prototype.connect = function () {
 
     connectionError.internal = error;
     self.state = 'error';
-    self.emitEvent('error', connectionError);
+    self.emitEvent('kerror', connectionError);
```
ballinette: `error` events have a special behaviour in Node, and throw an error if it is triggered with no listener (see https://nodejs.org/api/events.html#events_error_events )
So we need another event name for Kuzzle errors to avoid bad behaviour.

Looking for a better name, any suggestion is welcome ;)

src/networkWrapper/wrappers/websocket.js
```diff
@@ -84,113 +75,34 @@ function WSNode(host, port, ssl) {
...
   /**
    * Fires the provided callback whenever a connection error is received
    * @param {function} callback
    */
   this.onConnectError = function (callback) {
-    this.listeners.error.push({
-      fn: callback,
-      keep: true
-    });
+    this.addListener('wserror', callback);
```
ballinette: `error` events have a special behaviour in Node, and throw an error if it is triggered with no listener (see https://nodejs.org/api/events.html#events_error_events )
So we need another event name for Kuzzle errors to avoid bad behaviour.

Looking for a better name, any suggestion is welcome ;)

ballinette:
NB: the unit tests are still ugly, but I need to clean them deeply in a new PR

src/Kuzzle.js
```diff
@@ -29,6 +30,7 @@ function Kuzzle (host, options, cb) {
   if (!(this instanceof Kuzzle)) {
     return new Kuzzle(host, options, cb);
   }
+  KuzzleEventEmitter.call(this);
```
scottinet: nitpicking: if it's not a constructor, then it should not start with a capital K

edit: I'm confused. A few lines below there is a `new KuzzleEventEmitter()`... so what is called here exactly?

src/Kuzzle.js
```diff
@@ -310,6 +296,28 @@ function Kuzzle (host, options, cb) {
...
+Kuzzle.prototype = new KuzzleEventEmitter();
+
+/**
+* Emit an event to all registered listeners
+* An event cannot be emitted multiple times before a timeout has been reached.
+*/
+Kuzzle.prototype.emit = function(eventName) {
+  var
+    now = Date.now(),
+    protectedEvent = this.protectedEvents && this.protectedEvents[eventName];
```
scottinet: nitpicking: it's unnecessary to verify the existence of `this.protectedEvents` as we're guaranteed to have this property set in any `Kuzzle` instance.

src/eventEmitter/index.js
```diff
@@ -0,0 +1,196 @@
+var EventEmitter;
```
scottinet: Looking from the entire code below, this global variable appears to have no use at all and should be removed.

src/eventEmitter/index.js
```diff
@@ -0,0 +1,196 @@
...
+  if (typeof window !== 'undefined') {
+    this._events = {};
+    this._onceEvents = {};
+  }
+  this.removeAllListeners();
+}
+
+if (typeof window === 'undefined') {
+  EventEmitter = require('events');
+  KuzzleEventEmitter.prototype = new EventEmitter();
```
scottinet: This line and the one above can be replaced by:

```js
KuzzleEventEmitter.prototype = new (require('events'))();
```

src/Kuzzle.js
ballinette: It's necessary because I nead to call the parent's contructor within Kuzzle's constructor
the line below only applies the parent's prototype to current one, but does not call the eventEmitter's constructor when we instantiate a `Kuzzle` object.

That said, it's the line below that was wrong (or at least not the better syntax). I've just fixed it and replaced it by:
```javascript
Kuzzle.prototype = Object.create(KuzzleEventEmitter.prototype);
```

src/eventEmitter/index.js
ballinette: Oh yes, I had an eslint error with 
```javascript
KuzzleEventEmitter.prototype = new require('events')();
```
but I didn't think to try with the parenthesis :/

src/Kuzzle.js
scottinet: Thanks

ballinette:
Hi @stafyniaksacha 
Can you review this PR again ?

ballinette:
signature for `addListener` method has also been changed in PHP and Android SDKs:
https://github.com/kuzzleio/sdk-android/pull/123
https://github.com/kuzzleio/sdk-php/pull/64

PR #200, opened 2017-03-30
Add kuzzle info in Document object

AnthonySendra:


AnthonySendra:
@scottinet totally agree with you. I thought it was a good idea to keep the same name between attribute returned by the API and the attribute returned by the SDK.
Do we have to open an issue on kuzzle side and let this PR in WIP?

PR #208, opened 2017-04-21
Adds a new replaceUser method to Security

samniisan:
Linked to https://github.com/kuzzleio/kuzzle-sdk/issues/21

### DIFF
--- a/src/security/Security.js
+++ b/src/security/Security.js
@@ -509,13 +509,25 @@ Security.prototype.createUser = function (id, content, options, cb) {
     options = null;
   }
 
-  if (options) {
-    action = options.replaceIfExist ? 'createOrReplaceUser' : 'createUser';
-  }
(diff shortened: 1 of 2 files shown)


src/security/Security.js
```diff
@@ -509,11 +509,47 @@ Security.prototype.createUser = function (id, content, options, cb) {
     options = null;
   }
 
-  if (options) {
-    action = options.replaceIfExist ? 'createOrReplaceUser' : 'createUser';
+  if (options && options.replaceIfExist) {
+    self.fetchUser(id, function (fetchError, fetchResult) {
+      if (fetchResult instanceof User && options && options.replaceIfExist === true) {
```
stafyniaksacha: We can reject the query here if `fetchUser` return a user and `options.replaceIfExist !== true` without querying kuzzle

samniisan: Indeed, PR updated

PR #209, opened 2017-04-28
Add Collection:documentExists & Document:exists methods

samniisan:


ballinette:
need to merge https://github.com/kuzzleio/kuzzle/pull/821 before this

PR #210, opened 2017-05-08
 Harmonize event emitters implementation

ballinette:
main `Kuzzle` SDK prototype and websocket wrapper implementation both used their own event listeners.

Now, both Kuzzle and websocket wrapper inherits from Node build-in `EventEmitter` class (https://nodejs.org/api/events.html), with a polyfilled implementation for browser.

## Breaking Changes
method `addListener` no longer returns an ID, to be compliant with NodeJS implementation ( https://nodejs.org/api/events.html#events_emitter_on_eventname_listener )
It now returns the Emitter object itself, so that it can be chained.

For each event, the listener callback function is used as the key in listeners list

## Documentation
https://github.com/kuzzleio/documentation/pull/172

### DIFF
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -295,8 +295,7 @@ function Kuzzle (host, options, cb) {
     });
   }
 }
-
-Kuzzle.prototype = new KuzzleEventEmitter();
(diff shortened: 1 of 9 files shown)


ballinette:
See discussion on #197 that were approved, but merged on the wrong branch

PR #217, opened 2017-05-12
Added all new credential related routes

jenow:
Creation of new actions in auth controller:
- [x] deleteMyCredentials
- [x] createMyCredentials
- [x] getMyCredentials
- [x] hasMyCredentials
- [x] updateMyCredentials
- [x] validateMyCredentials
Creation of new actions in security controller:
- [x] createCredentials
- [x] deleteCredentials
- [x] getCredentialFields
- [x] getCredentials
- [x] getAllCredentialFields
- [x] hasCredentials
- [x] updateCredentials
- [x] validateCredentials

### DIFF
--- a/src/security/Security.js
+++ b/src/security/Security.js
@@ -751,4 +751,197 @@ Security.prototype.getUserRights = function (userId, options, cb) {
   });
 };
 
+/**
+ * Create credentials of the specified <strategy> for the user <kuid>.
+ *
+ * @param strategy
+ * @param kuid
+ * @param credentials
+ * @param options
+ * @param cb
+ * @returns {Security}
+ */
+Security.prototype.createCredentials = function (strategy, kuid, credentials, options, cb) {
+  if (!cb && typeof options === 'function') {
+    cb = options;
+    options = null;
+  }
+
+  this.kuzzle.query({controller: 'security', action: 'createCredentials'}, {_id: kuid, strategy: strategy, body: credentials}, options, function(err, res) {
+    if (!err) {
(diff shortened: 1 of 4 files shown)


src/security/Security.js
```diff
@@ -751,4 +751,197 @@ Security.prototype.getUserRights = function (userId, options, cb) {
...
+ * Create credentials of the specified <strategy> for the user <kuid>.
+ *
+ * @param strategy
+ * @param kuid
+ * @param credentials
+ * @param options
+ * @param cb
+ * @returns {Security}
+ */
+Security.prototype.createCredentials = function (strategy, kuid, credentials, options, cb) {
```
xbill82: What if `!strategy`?

src/security/Security.js
```diff
@@ -751,4 +751,197 @@ Security.prototype.getUserRights = function (userId, options, cb) {
...
+/**
+ * Delete credentials of the specified <strategy> for the user <kuid> .
+ *
+ * @param strategy
+ * @param kuid
+ * @param options
+ * @param cb
+ * @returns {Security}
+ */
+Security.prototype.deleteCredentials = function (strategy, kuid, options, cb) {
```
xbill82: What if `!strategy`?

xbill82: This applies to all the following methods that accept `strategy` as the first argument.

jenow: Then kuzzle wil answer an error

xbill82: ok then we assume that we leave the responsibility to Kuzzle. Approving.

PR #224, opened 2017-05-30
Add support for Document:m* methods

samniisan:
# Introduces
- Document;m* methods support (related to https://github.com/kuzzleio/kuzzle/pull/616)
- Corresponding unit tests

### DIFF
--- a/src/Collection.js
+++ b/src/Collection.js
@@ -382,6 +382,236 @@ Collection.prototype.getMapping = function (options, cb) {
 };
 
 /**
+ * Create the provided documents
+ *
+ * @param {Array.<document>} documents - Array of documents to create
+ * @param {object} [options] - Optional parameters
+ * @param {responseCallback} cb - Returns an instantiated CollectionMapping object
+ * @returns {object} this
+ */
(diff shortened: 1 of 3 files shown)


src/Collection.js
```diff
@@ -382,6 +382,236 @@ Collection.prototype.getMapping = function (options, cb) {
...
+ */
+Collection.prototype.mCreate = function (documents, options, cb) {
+  var
+    data = {
+      body: {},
+    },
+    self = this;
+
+  if (!Array.isArray(documents)) {
+    throw new Error('Collection.mCreate: documents parameter format is invalid (should be an array of documents)');
```
benoitvidis: open question: would it be simpler to send this error in the callback? Else, the end-user will have to write somehting like
```js
try {
  collection.mcreate(docs, function (error, response) {
    if (error) {
      // 1st error handling
    }
  });
}
catch (e) {
  // 2nd error handling
}

samniisan: Indeed, I will update this and let you know :+1:

samniisan: Done :)

PR #226, opened 2017-06-01
Network error handling 

stafyniaksacha:
## Bugfixes
- SocketIO: handle network errors (like we do in Websocket protocol wrapper)
- Websocket: arguments can differs from NodeJS and Browsers
- Auto resubscribe when a reconnection occurs after a network error


src/networkWrapper/wrappers/socketio.js
```diff
@@ -12,11 +22,75 @@ function SocketIO(host, port, ssl) {
...
+    var self = this;
+
     this.socket = window.io((this.ssl ? 'https://' : 'http://') + this.host + ':' + this.port, {
       reconnection: autoReconnect,
       reconnectionDelay: reconnectionDelay,
       forceNew: true
     });
+
+    this.socket.on('connect', function() {
+      if (self.wasConnected) {
```
benoitvidis: According to socket.io documentation, `connect` event is fired "upon a connection including a successful reconnection".
If this is the case, don't we have a double trigger with the `on('reconnect',..)` one?

src/networkWrapper/wrappers/socketio.js
```diff
@@ -12,11 +22,75 @@ function SocketIO(host, port, ssl) {
...
+        handler(error);
+      });
+
+      if (autoReconnect && !self.retrying) {
+        self.retrying = true;
+        setTimeout(function () {
+          self.retrying = false;
+          self.connect(autoReconnect, reconnectionDelay);
+        }, reconnectionDelay);
+      }
```
benoitvidis: isn't this part supposed to be natively handled by socket.io `reconnect` options?

ballinette:
missing description of what is the issue this PR fixes ?

src/networkWrapper/wrappers/socketio.js
stafyniaksacha: Nope, this is different cases:
- reconnection on `connect` event is used when proxy close connection
- reconnection on `reconnect` event is used when socket hang out and come back again via socket io reconnection logic

src/networkWrapper/wrappers/socketio.js
stafyniaksacha: I don't know why but reconnection from socketio does not work... (love socketio <3)

stafyniaksacha:
@ballinette @benoitvidis sorry for the description, i wasn't saved it ^^

src/Room.js
```diff
@@ -212,7 +213,16 @@ Room.prototype.renew = function (filters, notificationCB, cb) {
   self.kuzzle.subscriptions.pending[self.id] = self;
 
   subscribeQuery.body = self.filters;
-  subscribeQuery = self.kuzzle.addHeaders(subscribeQuery, this.headers);
+  subscribeQuery = self.kuzzle.addHeaders(subscribeQuery, self.headers);
+
+  networkListener = function() {
```
ballinette: I don't like this behaviour (too many "addListener" and "removeListener").

Better just add following lines into `src/Kuzzle.js` (at line [365](https://github.com/kuzzleio/sdk-javascript/blob/321211f442a14dad8136a9ed4d3dee112905d63f/src/Kuzzle.js#L365)) :
```javascript
self.network.onConnectError(function (error) {
// (...)
  disableAllSubscriptions.call(self);
// (...)
});

function disableAllSubscriptions() {
  var self = this;

  Object.keys(self.subscriptions).forEach(function (roomId) {
    Object.keys(self.subscriptions[roomId]).forEach(function (subscriptionId) {
      var subscription = self.subscriptions[roomId][subscriptionId];
      subscription.subscribing = false;
    });
  });
}
```

It should work, but without a complex listener management.

stafyniaksacha: Work like a charm !

src/networkWrapper/wrappers/socketio.js
benoitvidis: Well, I see that you intend to use them differently but this is not what the doc says. `connect` is fired each time a connection is made, including socket.io reconnections.

src/networkWrapper/wrappers/socketio.js
benoitvidis: Hum.. If this is the case, shouldn't we then just not use it at all? 
If it happens to work as expected, I still see an double trigger issue btw the connect and reconnect handlers (latest never being fired).

src/networkWrapper/wrappers/socketio.js
stafyniaksacha: Ok, so I removed `reconnect` listener

PR #227, opened 2017-06-02
Update MemoryStorage.*scan return value format

scottinet:
# Description

Following the work done in https://github.com/kuzzleio/documentation/pull/281, the returned values of `MemoryStorage.*scan` methods has been changed.

This PR updates the SDK to make it match the new return value format for these methods.

**Note:** this will trigger a patch version update on this SDK (5.0.2), as the published 5.x version is not usable until Kuzzle RC10 is out

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-sdk",
-  "version": "5.0.1",
+  "version": "5.0.2",
   "description": "Official Javascript SDK for Kuzzle",
   "author": "The Kuzzle Team <support@kuzzle.io>",
   "repository": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-sdk",
-  "version": "5.0.1",
+  "version": "5.0.2",
```
ballinette: I thought that package.json should be updated by the release PR, not there ?

scottinet: Yes, but as I stated last week, I plan to merge this PR directly on `master` once it is merged on `5.x`, and to rewrite the tag to make the `5.x` version of this SDK start at 5.0.2 instead of 5.0.0.
So I won't use the release tool and I want to merge this PR as-is, without modifying it before reaching master.

The idea is to avoid major versions jumps, as the `5.x` version has to be released before Kuzzle RC10. It is currently unusable anyway, until RC10 is released.

ballinette: ok

PR #232, opened 2017-06-16
Add SDK version into volatile

samniisan:
- Adds the current version of the SDK into volatile for every Kuzzle query.
- Modified webpack config to integrate the version at build time.

### DIFF
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -62,7 +62,7 @@ function Kuzzle (host, options, cb) {
         'queryError',
         'discarded'
       ],
-      writeable: false
+      writable: false
     },
     queuing: {
       value: false,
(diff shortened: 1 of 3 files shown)


src/Kuzzle.js
```diff
@@ -198,6 +198,10 @@ function Kuzzle (host, options, cb) {
       value: undefined,
       enumerable: true,
       writable: true
+    },
+    sdkVersion: {
+      value: (typeof SDKVERSION === 'undefined') ? JSON.parse(require('fs').readFileSync('package.json', {encoding: 'utf-8'})).version : SDKVERSION,
```
scottinet: You can require a JSON file directly:

`value: (typeof SDKVERSION === 'undefined') ? require('../package.json').version : SDKVERSION`

samniisan: Done

PR #233, opened 2017-06-16
Allow paginated search using ES's search_after in SearchResult:fetchNext

samniisan:
## Introduces
- Way to use ES's search_after if conditions are met when calling SearchResult:fetchNext (better than using scrolls, bypasses the default search limit count)

## Related issue
https://github.com/kuzzleio/kuzzle/issues/856

## Boyscoot
- Updates README.md (typos / various)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ You can access the Kuzzle repository on [Github](https://github.com/kuzzleio/kuz
 
 ## Basic usage
 
-Follow [Kuzzle Guide](http://docs.kuzzle.io/guide/getting-started/#sdk-play-time)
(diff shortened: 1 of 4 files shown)


benoitvidis:
@scottinet I agree on the scroll precedence that should not be overridden but to me, the user already has a way to specify he wants to use `scroll` over `search_after`, just by setting the `scroll` parameter in his search query.

scottinet:
Simple enough: this means `search_after` is only applied if `scrollId` is not set.

This requires some documentation though, but other than that, that's fine by me.

samniisan:
PR updated, scroll is now prioritized as long as the user specifies a scroll param.

PR #238, opened 2017-07-04
Login: move "expiresIn" option from request.body to request root attribute

ballinette:
fix https://github.com/kuzzleio/kuzzle-sdk/issues/39


src/Kuzzle.js
```diff
@@ -523,11 +526,11 @@ Kuzzle.prototype.login = function (strategy) {
 
   if (typeof credentials === 'object') {
     Object.keys(credentials).forEach(function (key) {
```
dbengsch: The body could just be set to credentials instead of copying it property by property

ballinette: done

PR #243, opened 2017-08-11
Refactor network wrappers

ballinette:
## Improvements

* Refactor the network wrappers to improve separation of concerns:
  + simplification of network states (`offline`, `connecting`, `connected` for realtime protocole ; `offline`, `ready` for future HTTP protocol)
  + each protocol is responsible of its states, and manage itself the offline mode
  + `Kuzzle` object still exposes queuing methods, but now delegates their implementation to network protocols.

* Refactor the Room management : the `Room` class now inherits from `EventEmitter`:
  + a room handles the subscription to Kuzzle API (one single subscription for each room, with its own filters and options)
  + listener callbacks can subscribe to a room to be notified by Kuzzle notifications.

Some sample codes to manage rooms:
```js
var 
  collection = kuzzle.collection('foo', 'bar'),
  filters = {equals: {foo: 'bar'}},
  room;

/*
* simple subscribe to new document with default scope/state options
*/
room = collection.subscribe(filters, function(data) {
  console.log('New document within the scope: ', data.document);
}).onDone(function(err, res) {
  if (err) {
    console.error('Error while subscribing to the room: ', err);
  } else {
    console.log('Subscription ready');
  }
});

/*
* Create a Room with custom options and add some listeners to it
*/

// create the room:
room = collection.room(filters, {state: all, scope: all, users: all, subscribeToSelf: false});

// listen to notifications about documents:
room.on('document', function(data) {
  if (data.scope === 'in') {
    console.log('New document within the scope: ', data.document);
  } else if (data.scope === 'out') {
    console.log('Document moved from the scope: ', data.document);
  } 
});

// listen to notifications about other users subscribing to the same room:
room.on('user', function(data) {
  if (data.user === 'in') {
    console.log('A user has joigned the room', data.volatile);
  } else if (data.user == 'out') {
    console.log('A user has leaved the room', data.volatile);
  }
  console.log('Number of listening users: ', data.result.count);
});

// subscribe to the room:
room.subscribe(function(err, res) {
  if (err) {
    console.error('Error while subscribing to the room: ', err);
  } else {
    console.log('Subscription ready');
  }
});
```

## Boyscout improvements
* use ES6 and Babel to browser transpilation (only for new and refactored classes for now)
* promisify somme missing method of `Kuzzle` and `Room` prototypes
* replace `jwtToken` by `jwt`
* improve unit tests

### Renamed properties and method in `Kuzzle` prototype
* `jwtToken` => `jwt`
* `getJwtToken()` => `getJwt()`
* `setJwtToken(token)' => `setJwt(token)`
* `unsetJwtToken()` => `unsetJwt()`
* `replayQueue()` => `playQueue()`

### Removal of `SubscribeResult` prototype:
* `onDone` method move to `Room` prototype.
* `Collection.subscribte()` method returns now the created `Room` object

### Changed signature for `Collection.room` method and `Room` constructor 

`collection.room([options])` => `collection.room(filters, [options])`
`new Room(collection, [options])` => `new Room(collection, filters, [options])` 

### Changed signature for `Room.renew` method

`renew([filters], notificationCallback, subscriptionCallback)` => `renew(subscriptionCallback)`

* `filters` is now an immutable property (we need to create a new Room if we want to change the filters)
* `notificationCallback` has become useless since the `Room` is now an event emitter:
```js
// Old code:
room.renew(notificationCallback, subscriptionCallback);

// ===> New Code:
room.renew(subscriptionCallback).addListener('document', notificationCalllback);
```

### New option `protocol` to Kuzzle constructor
Must be filled to use another protocol than native websocket
currently, 2 supported values: 
* `websocket` (default)
* `socketio`
:warning: The network wrappers does not automatically fallback to socketIO anymore if Websocket is not supported.


src/Collection.js
```diff
@@ -817,10 +812,6 @@ Collection.prototype.scrollSpecifications = function (scrollId, options, cb) {
 
   this.kuzzle.callbackRequired('Collection.scrollSpecifications', cb);
 
-  if (options && options.scroll) {
```
ballinette: not needed here since `scroll` option is already managed within `kuzzle.query`method

src/Collection.js
```diff
@@ -758,10 +757,6 @@ Collection.prototype.scroll = function (scrollId, options, filters, cb) {
 
   request.scrollId = scrollId;
 
-  if (options && options.scroll) {
```
ballinette: not needed here since `scroll` option is already managed within `kuzzle.query`method

src/eventEmitter/index.js
```diff
@@ -129,7 +129,7 @@ if (typeof window === 'undefined') {
     listeners = this._events && this._events[eventName];
 
     if (!listeners || !listeners.length) {
-      return;
+      return false;
```
ballinette: fix KuzzleEventEmitter methods signatures, to be consistant with NodeJS `event` specitications:

emit should return a boolean value.
(see https://nodejs.org/api/events.html#events_emitter_emit_eventname_args )

src/networkWrapper/wrappers/abstract/realtime.js
```diff
@@ -0,0 +1,389 @@
...
+  }
+
+  dequeuingProcess();
+}
+
+/**
+ * Clean history from requests made more than 10s ago
+ */
+function cleanHistory (requestHistory) {
+  var
```
scottinet: Nitpicking: consider using `const` instead of `var`.

src/networkWrapper/wrappers/abstract/realtime.js
```diff
@@ -0,0 +1,389 @@
...
+    now = Date.now();
+
+  Object.keys(requestHistory).forEach(function (key) {
```
scottinet: Nitpicking: replace with `for(const ... of ...)`

src/networkWrapper/wrappers/websocket.js
```diff
@@ -1,55 +1,44 @@
...
+      this.clientConnected();
     };
 
-    this.client.onclose = function (closeEvent, message) {
-      var error;
-      var status;
-      var reason = message;
+    this.client.onclose = (closeEvent, message) => {
+      let
+        error,
```
scottinet: This `error` variable is shadowed by the one declared in the function starting line 65.

Suggestion: declare this variable as a `const` directly line 58, as it seems to be the only block where it is used.

src/networkWrapper/wrappers/websocket.js
```diff
@@ -63,113 +52,60 @@ function WSNode(host, port, ssl) {
...
 
-        onClientNetworkError(self, autoReconnect, reconnectionDelay, error);
+        this.clientNetworkError(error);
       }
     };
 
-    this.client.onerror = function (error) {
+    this.client.onerror = error => {
       if (!(error instanceof Error)) {
         error = new Error(error);
```
scottinet: Nitpicking: function parameters should not be overwritten.

src/Kuzzle.js
```diff
@@ -239,6 +115,81 @@ function Kuzzle (host, options, cb) {
...
+            delete data.result;
+          }
+          room.notify(data);
+        };
+
+      if (this.jwt !== undefined) {
+        object.jwt = this.jwt;
+      }
+
+      if (room.volatile) {
```
ballinette: done

test/kuzzle/connect.test.js
```diff
@@ -28,310 +32,142 @@ describe('Kuzzle connect', function () {
...
-      process.nextTick(function () {
-        should(renewStub).be.calledOnce();
-        done();
-      });
+      clock.tick();
+      should(eventStub).be.calledOnce();
     });
 
-    it('should dequeue requests automatically on a connection success', function (done) {
+    it('should keep a valid JWT Token at reconnection', function () {
```
dbengsch: a valid JWT ?

test/kuzzle/connect.test.js
```diff
@@ -28,310 +32,142 @@ describe('Kuzzle connect', function () {
...
-        done();
-      });
+      should(kuzzle.jwt).be.eql('foobar');
+      should(eventStub).not.be.called();
     });
-  });
 
-  describe('=> on reconnection', function () {
-    it('should exit offline mode when reconnecting', function (done) {
+    it('should empty the JWT Token at reconnection if it has expired', function () {
```
dbengsch: JWT ?

src/networkWrapper/index.js
```diff
@@ -1,27 +1,26 @@
...
-    throw new Error('Aborting: no websocket support detected and no socket.io library loaded either.');
+function network(protocol, host, options) {
+  switch (protocol) {
+    case 'websocket':
+      if (typeof window !== 'undefined' && typeof WebSocket === 'undefined') {
+        throw new Error('Aborting: no websocket support detected.');
+      }
+      return new (require('./protocols/websocket'))(host, options);
+    case 'socketio':
+      if (!window.io) {
```
xbill82: We assume that no one will ever try to use Socket.IO outside a browser?
(after all, sounds fair to me, just asking to be sure it's on purpose)

test/kuzzle/methods.test.js
```diff
@@ -376,344 +315,467 @@ describe('Kuzzle methods', function () {
...
-  describe('#setDefaultIndex', function () {
-    it('should throw an error if the provided index is not a string', function () {
-      should((function () {kuzzle.setDefaultIndex();})).throw();
-      should((function () {kuzzle.setDefaultIndex({});})).throw();
-      should((function () {kuzzle.setDefaultIndex([]);})).throw();
-      should((function () {kuzzle.setDefaultIndex(123);})).throw();
-      should((function () {kuzzle.setDefaultIndex(null);})).throw();
-      should((function () {kuzzle.setDefaultIndex(undefined);})).throw();
+  describe('#getJwt', function () {
+    it('should return the current jwt token', function () {
```
dbengsch: JWT

test/kuzzle/methods.test.js
```diff
@@ -976,142 +916,112 @@ describe('Kuzzle methods', function () {
...
+      kuzzle.setJwt({foo: 'bar'});
 
-        should(args[0].controller).be.exactly('auth');
-        should(args[0].action).be.exactly('getMyCredentials');
-        should(args[2]).be.exactly(null);
-        done();
-      });
+      should(kuzzle.jwt).be.undefined();
+      should(loginAttemptStub).be.calledOnce();
+      should(loginAttemptStub).be.calledWith({ error: 'Cannot find a valid JWT token in the following object: {"foo":"bar"}', success: false });
```
dbengsch: :(

src/networkWrapper/index.js
ballinette: If we are outside a browser, we have native websocket support, so I don't see any reason to use socketIO...

PR #245, opened 2017-09-18
Make collection.create() pass the collection object to the callback

xbill82:
The method `kuzzle.collection('foo').create()` passed to the callback the response from the server. We estimate that passing the collection makes the API easier to use. It notably allows to do

```javascript
kuzzle
  .collection('bar')
  .createPromise()
  .then((collection) => {
    return collection
       .createDocumentPromise(...)
  })
```

## Boyscouting

Included `kuzzle.createIndex` in the promisification whitelist.


scottinet:
This needs an update to the SDK documentation and to other SDKs as well

test/network/networkWrapper.test.js
```diff
@@ -10,8 +10,10 @@ var
 
 describe('Network Wrapper', function () {
   afterEach(function () {
-    window = undefined; // eslint-disable-line
-    WebSocket = undefined; // eslint-disable-line
+    window = undefined;
+    /*eslint no-native-reassign: "off"*/
```
scottinet: Nitpicking: this disables the rule for the entire file. Prefer the following:
`// eslint-disable-next-line no-native-reassign, no-global-assign`

PR #246, opened 2017-09-19
Better instructions to use in the browser

xbill82:


README.md
```diff
@@ -59,22 +59,39 @@ This SDK can be used either in NodeJS or in a browser.
 npm install kuzzle-sdk --save
 ```
 
-### Javascript
-
-Clone this github repository and run ``npm run build``. A ``dist`` directory will be created, containing a browser version of this SDK.
+### Browser
 
+To run the SDK in the browser, you need to pick the built version.
+Clone this github repository and run `npm run build`. A `dist` directory will be created, containing a browser version of this SDK.
```
ballinette: or simply download https://github.com/kuzzleio/sdk-javascript/blob/master/dist/kuzzle.js
(but ihmo we should put dist/ away from the source code... to be discussed with the whole team...)

PR #249, opened 2017-09-27
Throw when using Nodejs version in browser

xbill82:
By injecting a `BUILT` constant via Webpack, and by checking the `window` global, we are able to detect if the user is using the Nodejs version in the browser.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,4 +5,12 @@ var
 // Adds on the fly methods promisification
 Kuzzle.prototype.bluebird = bluebird;
 
+if (typeof window !== 'undefined' && typeof BUILT === 'undefined') {
+  /* eslint-disable no-console */
+  console.warn('It looks like you are using the Nodejs version of Kuzzle SDK ' +
+               'in a browser. ' +
+               'It is strongly recommended to use the browser-specific build instead. ' +
+               'Learn more at https://github.com/kuzzleio/sdk-javascript/tree/master#javascript');
+}
+
 module.exports = Kuzzle;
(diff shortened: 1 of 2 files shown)


ballinette:
I suggest to be more strict and throw an error if we try to use the nodejs version in a browser

xbill82:
@ballinette Why? Technically, the SDK works. It is a common practice in famous browser libraries like React, Redux, Vuejs, etc... To display warnings when something _looks_ bad, but maybe the user just knows what he/she is doing.

ballinette:
@xbill82 so I don't understand the goal here.
If the SDK works perfectly with the Node version in the brother, why should we display a warning message ?
If not, why do we allow to use it ?

Is there any acceptable use case that a browser app could use the nodejs version of the sdk instead of the built one ?

index.js
```diff
@@ -5,4 +5,11 @@ var
 // Adds on the fly methods promisification
 Kuzzle.prototype.bluebird = bluebird;
 
+if (typeof window !== 'undefined' && typeof BUILT === 'undefined') {
+  throw new Error('It looks like you are using the Nodejs version of Kuzzle SDK ' +
+               'in a browser. ' +
+               'It is strongly recommended to use the browser-specific build instead. ' +
```
scottinet: Since we now throw an error, we are more than "strongly recommending" to use the built-version of the SDK. :-)

index.js
```diff
@@ -5,4 +5,11 @@ var
...
+               'Learn more at https://github.com/kuzzleio/sdk-javascript/tree/master#browser');
```
scottinet: I cannot find that anchor in our README file, either in `master`, `5.x` or `6.x`

ballinette: the anchor in the README is `#javascript` 
should rather be `#browser`, imho, so the link here is correct, bug we need to fix the Readme.

xbill82: @scottinet https://github.com/kuzzleio/sdk-javascript/tree/5.x#browser

xbill82:
Merging as 4 :white_check_mark:

PR #263, opened 2017-12-05
return result in case of partial errors for mupdate/mcreate/mdelete methods

ballinette:


src/Collection.js
```diff
@@ -367,7 +367,7 @@ Collection.prototype.mCreateDocument = function (documents, options, cb) {
   });
 
   self.kuzzle.query(this.buildQueryArgs('document', 'mCreate'), data, options, cb && function (err, res) {
-    cb(err, err ? undefined : res.result);
+    cb(err, res && res.result);
```
scottinet: This will invoke the callback with a `false` result part instead of a `undefined` one in case of a non-partial error.

Fix: `cb(err, res ? res. result : undefined);`

src/Collection.js
```diff
@@ -403,7 +403,7 @@ Collection.prototype.mCreateOrReplaceDocument = function (documents, options, cb
   });
 
   self.kuzzle.query(this.buildQueryArgs('document', 'mCreateOrReplace'), data, options, cb && function (err, res) {
-    cb(err, err ? undefined : res.result);
+    cb(err, res && res.result);
```
scottinet: Wrong result value in case of error (see above)

src/Collection.js
```diff
@@ -437,7 +437,7 @@ Collection.prototype.mDeleteDocument = function (documentIds, options, cb) {
   self.kuzzle.callbackRequired('Collection.mDelete', cb);
 
   self.kuzzle.query(this.buildQueryArgs('document', 'mDelete'), data, options, cb && function (err, res) {
-    cb(err, err ? undefined : res.result);
+    cb(err, res && res.result);
```
scottinet: Wrong result value in case of error (see above)

src/Collection.js
```diff
@@ -540,7 +540,7 @@ Collection.prototype.mUpdateDocument = function (documents, options, cb) {
   });
 
   self.kuzzle.query(this.buildQueryArgs('document', 'mUpdate'), data, options, cb && function (err, res) {
-    cb(err, err ? undefined : res.result);
+    cb(err, res && res.result);
```
scottinet: Wrong result value in case of error (see above)

PR #265, opened 2017-12-07
Custom event emitter

scottinet:
# Description

* Remove the possibility of having duplicate listeners added on a same event. Duplicates are now silently discarded instead of being added. To counterbalance this, event emitter methods throw only on rare occasions
* add the `addOnceListener` method, as it's more in line with other method names than `once`, which is now an alias (see below about aliases)

# Other related changes

* since we have rewritten all methods from NodeJS' event emitter class, there is no point in extending it, so I've removed `window` manipulation and the need to require `events` when on NodeJS side
* the `KuzzleEventListener` object is now an ES6 class (added to babel configuration for conversion to ES5)
* simplify the way `once` listeners are handled
* make sure functions returned values are consistent 
* asking for a list of listeners on an unknown event won't create an empty array (potential memleak, albeit small)
* the socketio network layer now correctly expose the whole event emitter API, with a bridge between our `KuzzleEventEmitter` class and socketio's dedicated event emitter 
* aliases: I think we do not want too many aliases, as this leads to [many questions](https://www.google.fr/search?client=ubuntu&channel=fs&q=difference+between+on+and+addlistener&ie=utf-8&oe=utf-8&gfe_rd=cr&dcr=0&ei=UxUpWpTSFdCA3gOOo79A) about the difference between a method and its corresponding alias, so I removed most of them:
  * `off` + `removeListener` => `removeListener` 
  * `emitEvent` + `emit` => `emit`
  * `on` and `once` are now aliases specific to our Javascript SDK implementation, as those two are very commonplace when it comes to event emitters in JS. This is (and should be) the only case where an alias exists

# Other changes

* convert Kuzzle object into an ES6 class


src/networkWrapper/protocols/socketio.js
```diff
@@ -49,36 +43,85 @@ class SocketIO extends RTWrapper {
...
+    return this;
+  }
+
+  on(event, callback) {
+    return this.addListener(event, callback);
+  }
+
+  prependListener(event, callback, once = false) {
+    this._addEventWrapper(event, callback, once);
+    return this.prependListener(event, callback, once);
```
ballinette: infinite loop ?

src/networkWrapper/protocols/socketio.js
```diff
@@ -49,36 +43,85 @@ class SocketIO extends RTWrapper {
...
+    return this;
   }
 
+  /**
+   * Unregisters all listeners either from an event, or from all events
+   *
+   * @param {string} [event]
+   */
+  removeAllListeners(event) {
+    if (event !== undefined && this.eventsWrapper[event] !== undefined) {
```
ballinette: if `event` is defined but `this.eventsWrapper[event]` is not, we will execute the `else` code, and so remove all listeners of all events ??

src/eventEmitter/index.js
```diff
@@ -1,186 +1,142 @@
...
+      this._events[eventName] = [new Listener(listener, once)].concat(this._events[eventName]);
+    }
 
     return this;
-  };
+  }
 
-  KuzzleEventEmitter.prototype.prependOnceListener = function(eventName, listener) {
-    var onceListeners;
+  addOnceListener (eventName, listener) {
```
ballinette: is it necessary to redefine these methods ?

src/networkWrapper/protocols/socketio.js
```diff
@@ -49,36 +43,85 @@ class SocketIO extends RTWrapper {
...
+   * Registers a callback on an event. Once 1 message is received, fires the
    * callback and unregister it afterward.
    *
-   * @param {string} roomId
+   * @param {string} event
    * @param {function} callback
    */
-  once(roomId, callback) {
-    this.socket.once(roomId, callback);
+  addOnceListener(event, callback) {
```
ballinette: is it necessary to redefine these methods ?

src/networkWrapper/protocols/socketio.js
scottinet: I replaced `super` by `this` as a last-minute change, I don't even know why :disappointed:

scottinet:
@ballinette > good catches :+1:

src/networkWrapper/protocols/socketio.js
scottinet: No, it is not. I removed it along with a few others (`on`, `prependOnceListener`, ...)

PR #271, opened 2018-02-23
HTTP Protocol suport

### LINKED ISSUE #167: Provide an option to use the HTTP protocol instead of the WebSocket one in the SDK
Are you going to add an option to use the HTTP protocol in the SDK ? It is currently using the websocket protocol and we don't need to maintain the connection between kuzzle and our client. 

It seems that there is some limitations in terms of performance (WS vs HTTP). 

We are currently using the REST API with Axios (HTTP client library) and we think it's best to use the SDK if you are going to develop this feature soon. We are waiting for this information to decide if we use the REST API with Axios or the javascript SDK.

ballinette:
Add support for HTTP protocol

fix #167
fix also #282


src/networkWrapper/protocols/abstract/common.js
```diff
@@ -0,0 +1,303 @@
...
+        error.status = response.status;
+        network.emit('queryError', error, request, cb);
+      }
+
+      if (cb) {
+        cb(error, response);
+      }
+    });
+  }
+  // Track requests made to allow Room.subscribeToSelf to work
```
scottinet: this comment seems to be a leftover from previous sdk versions

PR #285, opened 2018-03-01
SDK Refactor: server controller

ballinette:
* Added Server controller with the following methods:
  - adminExists()
  - getAllStats()
  - getConfig()
  - getLastStats()
  - getStats()
  - info()
  - now()

* Don't use callbacks anymore for new methods (use bluebird promises for both nodejs and browsers)
NB: for the timebeeing, the new methods use `kuzzle.queryPromise` to keep the code compatible avec methods that are still not refactored.
The `kuzzle.query` method refactoring will be finalize at last step of the related EPIC (see https://github.com/kuzzleio/kuzzle-backlog/issues/112 )


ballinette:
@benoitvidis As explained in the PR's description, I let the bluebird's promisification for `query` here only to keep the whole code working until the refactoring is not totally finished yet:

If we modify first the `query` method, all methods that use it will be broken. So we have 2 options:
* either have a broken code until we have rewriten all the SDK (and be sure we won't draft any new release before that)
* or keep `query` as is for now and rewrite it at the final step of the refactoring.

I personally chose the 2nd option because IMHO, it's a better way for me to make sure each PR don't break methods that are still not rewritten...

ballinette:
@benoitvidis 
Ok, since #288 has been merged and already contains the refactor of `query` method, I moved to it in this one, and resovled merge conflicts.

@jenow you may want to review this PR again.

PR #286, opened 2018-03-02
volatile data can also be null

scottinet:
When looking if a `sdkInstance` identifier is present in volatile data, we first check whether volatiles are defined or not. But they can also be null, leading to a sdk crash when that happens.
This PR fixes that issue.


src/networkWrapper/protocols/abstract/realtime.js
```diff
@@ -212,7 +212,7 @@ class RTWrapper extends KuzzleEventEmitter {
         return cb(error);
       }
       this.on(response.result.channel, data => {
-        data.fromSelf = data.volatile !== undefined && data.volatile.sdkInstanceId === this.id;
+        data.fromSelf = data.volatile !== undefined && data.volatile !== null && data.volatile.sdkInstanceId === this.id;
```
benoitvidis: is there any chance `data.volatile`  can be a scalar?

scottinet: No, the `RequestInput` object forbids it: https://github.com/kuzzleio/kuzzle-common-objects/blob/master/lib/models/requestInput.js#L249

benoitvidis: ok, great

PR #288, opened 2018-03-13
SDK refactor: Auth controller

jenow:
Added the auth controller with the following method:
- checkToken
- createMyCredentials
- credentialsExist
- deleteMyCredentials
- getCurrentUser
- getMyCredentials
- getMyRights
- getStrategies
- login
- logout
- updateMyCredentials
- User updateSelf
- validateMyCredentials

Merge the branch feature/query-pormise so query returns promise


ballinette:
Note: to help the code review and concentrate only on `Auth` controller, the `feature/query-pormise` branch should have its own PR.
Or, this should be delayed at the last step of the refactoring process (see my comment here: https://github.com/kuzzleio/sdk-javascript/pull/285#issuecomment-372671542

src/Auth.js
```diff
@@ -0,0 +1,224 @@
...
+  /**
+   * Send login request to kuzzle with credentials
+   * If login success, store the jwt into kuzzle object
+   *
+   * @param strategy
+   * @param credentials
+   * @param expiresIn
+   * @returns {Promise|*|PromiseLike<T>|Promise<T>}
+   */
+  login(strategy, ...args) {
```
ballinette: We should avoid implicit arguments, and use better an explicit signature:
`login(strategy, credentials, expiresIn)`
If we need to call `login` without credentials, just call `login(strategy, null, expiresIn)`

src/Auth.js
```diff
@@ -0,0 +1,220 @@
...
+   */
+  login(strategy, credentials, expiresIn) {
+    if (!strategy || typeof strategy !== 'string') {
+      return Promise.reject(new Error('Auth.login: strategy required'));
+    }
+
+    const
+      request = {
+        strategy,
+        body: {}
```
ballinette: `request.body` should contain content from `credentials` argument

Issue #298, opened 2018-07-14
Suggestion: Event `disconnected` should be fired when .disconnect is called


lethak:
## Suggested Behavior
Event `disconnected` should be triggered upon manually calling `.disconnect()` in manual connection mode.

Perhaps with a callback argument specifying is was from a manual action and not an unexpected one.

## Current Behavior
Event `disconnected` is NOT triggered upon manually calling `.disconnect()`

## Steps to Reproduce
```javascript
const vm = {isConnected:false}
kuzzle.addListener('connected', () => {
  vm.isConnected = true
})
kuzzle.addListener('reconnected', () => {
  vm.isConnected = true
})
kuzzle.addListener('disconnected', () => {
  vm.isConnected = false
})

kuzzle.connect() //  event connected is triggered, vm.isConnected === true

// ... do some stuff ...

kuzzle.disconnect() //  event disconnected is NOT triggered, vm.isConnected === true and kuzzle.state === 'disconnected'
```

## Context
Trying to create a Vue JS wrapper
Using docker image: kuzzleio/kuzzle:1.3.1

Aschen:
Hi @lethak !
Thanks for your report, we will investigate this issue shortly.

Aschen:
Hi @lethak,

This behavior is intentional. The `disconnected` event is triggered by unexpected disconnection only, so you can implement some logic to reconnect to Kuzzle. In the meantime, the request will be queued.  
 
In the other hand, if you are manually call the `disconnect()` method, the disconnection in intentional. Kuzzle switches to `disconnected` state. Request won't be queued and you can't make any use of the API until a call to `connect()`.  

We are aware that the documentation was not very clear about this, we have now improved it.  

Thanks you for your feedback !

lethak:
Hi,

I understood that from the documentation, but this is a _suggestion_ to change this frankly illogical behavior.

Of course I can create layers upon layers of my own event hub in front of kuzzle, but this is supposed to be a core event imho.

Aschen:
We agreed that the name `disconnect()` for this method may be inappropriate and can lead to confusion. 
We are thinking about something else for the next version, feel free to propose a different name :)

scottinet:
Hi @lethak,

This is a choice we made deliberately, because the SDK instance is in an entirely different state whether the event is called after an unexpected disconnection, or after an explicit call to `disconnect()`. 
Listeners on that event would have to handle different cases, and it kinda kills the purpose of an event system.

[...]

PR #299, opened 2018-07-16
Fix disconnected event

Aschen:
## What does this PR do?

This PR fix a bug with the `disconnected` event that wasn't fired when using the method `disconnect()`.

Fix https://github.com/kuzzleio/sdk-javascript/issues/298

  - Step 1 : Execute this code at sdk root : https://gist.github.com/Aschen/daec7eded0c5d9f02620b0d289c36395
  - Step 2 : You should see `disconnected` on standard output

### DIFF
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -1146,6 +1146,8 @@ Kuzzle.prototype.disconnect = function () {
       delete this.collections[collection];
     }
   }
+
+  this.emit('disconnected');
 };
 
 /**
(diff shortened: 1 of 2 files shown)


Aschen:
For me there is a naming problem on the `5.x` version between the `disconnect()` method which manually close the connection to Kuzzle and put the SDK in `disconnected` mode (no more queing etc) and the `disconnected` event which corresponds more to an `unexpectedDisconnection` event (the state change to `offline` and request are queued).

I just checked the `6.x` branch, in the next version of the SDK a call to `disconnect()` put the SDK in [offline mode](https://github.com/kuzzleio/sdk-javascript/blob/6.x/src/networkWrapper/protocols/websocket.js#L114) but here is still no `disconnected` event fired.   
That suits me, a manual call disconnection should not allow to execute the same business logic as an unexpected disconnection.

Aschen:
Documentation PR :  https://github.com/kuzzleio/documentation/pull/495

PR #319, opened 2018-10-06
Implementing support for Typescript

ScreamZ:
# What does this PR do?
<p align="center">
Typescript Logo
</p>

This pull request aims to add typescript support to the javascript SDK. Thus, auto-completion will be available on Typescript compliant IDE for both `.js` and `.ts` files. Also compilation typescheking on typescript based project.

----

Nowadays there are many users are that are using Typescript, especially for production apps with high business risks. Shipping the definitions file in the library can encourage new users and experimented one to experience Kuzzle due to the smooth development experience provided.

1. Create a new project and npm install the new sdk. This can be a Typescript project or a simple js project, the main difference is that for a vanilla project you'll only get auto-completion and you won't be able to use provided types to your own code.
2. Import as you would normally do.
3. Check auto-completion and compilations warning (Vscode or Intellij software work well)

### Boyscout

In the new documentation Kuzzle `options` parameter is marked as required but you provided a default value. This is not semantically correct, I also added a check for this.

# Documentation feedback

A thread is describing documentation lacks that results of some todos in this PR. https://github.com/kuzzleio/documentation-V2/issues/199

Thanks for reviewing,
Regards

### DIFF
--- a/kuzzle-sdk.d.ts
+++ b/kuzzle-sdk.d.ts
@@ -62,7 +62,7 @@ declare module "kuzzle-sdk" {
             sslConnection?: boolean;
         })
     }
-    export class Websocket implements KuzzleAbstractProtocol {
+    export class WebSocket implements KuzzleAbstractProtocol {
         public readonly name: string;
         public readonly autoReconnect: boolean
         public readonly reconnectionDelay: number
--- a/kuzzle-sdk.d.ts
+++ b/kuzzle-sdk.d.ts
@@ -1,0 +1,336 @@
+import { resolve } from "dns";
+
+/**
+ * Type definitions for Kuzzle javascript SDK.
+ * 
+ * Project: Kuzzle Javascript SDK - https://github.com/kuzzleio/sdk-javascript
+ * Definitions by: Andréas 'ScreamZ' HANSS <andreas.hanss67@gmail.com>
+ * ---
+ * Github: https://github.com/ScreamZ
+ * LinkedIn: https://www.linkedin.com/in/andreas-hanss/
+ * 
+ * If you find any issues, feel free to report or make a pull request.
+ */
+declare module "kuzzle-sdk" {
+    import { Explanation, Refresh } from "elasticsearch"
+
+    export class Kuzzle {
+        public autoQueue: boolean;
+        public autoReplay: boolean;
+        public autoResubscribe: boolean;
+        public jwt: string;
+        public offlineQueueLoader: () => Array<{ query: KuzzleRequest, resolve: (response: KuzzleResponse) => Promise<any>, reject: (error: KuzzleError) => Promise<any> }>;
+        public queueFilter: (request: KuzzleRequest) => boolean
+        public queueMaxSize: number;
+        public queueTTL: number;
+        public replayInterval: number;
+        public volatile: ObjectWithAnyKeys;
+        public readonly offlineQueue: ObjectWithAnyKeys[]
+        public readonly protocol: KuzzleAbstractProtocol
+
+        constructor(protocol: KuzzleAbstractProtocol, options: KuzzleOptions);
+
+        // Core
+        addListener(event: KuzzleEvent, handler: (event: any) => void): Kuzzle;
+        connect(): Promise<void>;
+        disconnect(): void;
+        flushQueue(): Kuzzle;
+        playQueue(): Kuzzle;
+        query(request: KuzzleRequest, options?: QueryOptions): KuzzleResponse
+        startQueuing(): Kuzzle;
+        stopQueuing(): Kuzzle;
+
+        // Modules
+        auth: KuzzleModules.Auth
+        bulk: KuzzleModules.Bulk
+        collection: KuzzleModules.Collection
+        document: KuzzleModules.Document
+        index: KuzzleModules.Index
+        ms: KuzzleModules.Ms
+        realtime: KuzzleModules.Realtime
+        server: KuzzleModules.Server
+    }
+
+    export interface KuzzleAbstractProtocol {
+        readonly name: string
+    }
+
+    export class Http implements KuzzleAbstractProtocol {
+        public readonly name: string;
+        constructor(host: string, options?: {
+            port?: number;
+            sslConnection?: boolean;
+        })
+    }
+    export class Websocket implements KuzzleAbstractProtocol {
+        public readonly name: string;
+        public readonly autoReconnect: boolean
+        public readonly reconnectionDelay: number
+
+        constructor(host: string, options?: {
+            port?: number;
+            sslConnection?: boolean;
+            autoReconnect?: boolean;
+            reconnectionDelay?: number;
+        })
+    }
+    export class SocketIO implements KuzzleAbstractProtocol {
+        public readonly name: string;
+        public readonly autoReconnect: boolean
+        public readonly reconnectionDelay: number
+
+        constructor(host: string, options?: {
+            port?: number;
+            sslConnection?: boolean;
+            autoReconnect?: boolean;
+            reconnectionDelay?: number;
+        })
+    }
+
+    type ObjectWithAnyKeys = {
+        [propName: string]: any;
+        [propName: number]: any;
+    }
+
+    interface QueryOptions extends ObjectWithAnyKeys {
+        queuable?: boolean;
+    }
+
+    interface RefreshableOptions extends QueryOptions {
+        refresh?: Refresh
+    }
+
+    interface RefreshRetryOptions extends RefreshableOptions {
+        retryOnConflict?: number;
+    }
+
+    interface PaginatorOptions extends QueryOptions {
+        from?: number;
+        size?: number;
+    }
+
+    interface ScrollOptions extends PaginatorOptions {
+        scroll?: string
+    }
+
+    type RealtimeScope = "in" | "out"
+    interface SubscribeOptions {
+        scope?: "all" | RealtimeScope | "none";
+        users?: "all" | RealtimeScope | "none";
+        subscribeToSelf?: boolean;
+        volatile?: ObjectWithAnyKeys
+    }
+
+    interface KuzzleOptions {
+        autoQueue?: boolean;
+        autoReplay?: boolean;
+        autoReconnect?: boolean;
+        autoResubscribe?: boolean;
+        eventTimeout?: number;
+        offlineMode?: "manual" | "auto";
+        queueTTL?: number;
+        queueMaxSize?: number;
+        replayInterval?: number;
+        volatile?: ObjectWithAnyKeys;
+    }
+
+    enum KuzzleEvent {
+        CONNECTED = "connected",
+        DISCARDED = "discarded",
+        DISCONNECTED = "disconnected",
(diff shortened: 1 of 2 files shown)


Aschen:
Hi !
Thanks you for your work, it's a great idea to have a Typescript definition file for IDE. This will please few team members I think :)

Aschen:
I will try to answer to your question:
 - About the `options` parameter in Kuzzle constructor: this parameter is optional because it depends on the Protocol used. For exemple the FunnelProtocol does not have `options` (See https://github.com/kuzzleio/kuzzle/blob/master/lib/api/core/sdk/funnelProtocol.js and https://github.com/kuzzleio/kuzzle/blob/master/lib/api/core/plugins/pluginContext.js#L71)
 - We use deconstructing assignment with `const { Kuzzle } = require('kuzzle-sdk');` because we also expose a `KuzzleAbstractNetwork` (see https://github.com/kuzzleio/sdk-javascript/pull/294)
 - And yes we didn't have documented the custom protocol object format ;)
 - `volatile` could be an object of any key
 - `KuzzleError` has message (mandatory), status (mandatory), stack (optional string) 
 - A `RawKuzzleResponse` represent the content of the response returned by the Kuzzle API, so it can vary
 - `index.list` and `index.mDelete` didn't return correctly, fixed here https://github.com/kuzzleio/sdk-javascript/pull/320

kuzzle-sdk.d.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    type EmbededProtocol = "http" | "websocket" | "socketio"
+    type CustomProtocol = object
+
+    interface QueuableOpts {
+        queuable: boolean;
+    }
+
+    interface KuzzleOptions {
```
Aschen: `KuzzleOptions` have some default key used by Kuzzle SDK itself but a custom protocol could accept other key for its configuration

kuzzle-sdk.d.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        // Modules
+        bulk: Bulk
+        index: Index
+    }
+
+    type EmbededProtocol = "http" | "websocket" | "socketio"
+    type CustomProtocol = object
+
+    interface QueuableOpts {
```
Aschen: It's not only `queuable`, a better name could be `QueryOptions`

src/Kuzzle.js
```diff
@@ -32,13 +32,17 @@ class Kuzzle extends KuzzleEventEmitter {
    * @param network - the network wrapper to use. if the argument is a string, creates an embeed network wrapper if a
    * @param [options] - Connection options
    */
-  constructor(network, options = {}) {
+  constructor(network, options) {
```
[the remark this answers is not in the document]
ScreamZ: @Aschen All right, the v2 documentation is saying the opposite.

Also `host` parameter has to be provided in options and is required. How can we make all of this coherent ?

kuzzle-sdk.d.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    interface RawKuzzleResponse {
+        _index: string,
+        _type: string,
+        _id: string,
+        _version: number,
+        result: 'created' | "", // TODO: 
+        _shards: { total: number, successful: number, failed: number },
+        created: boolean,
+        _source: any // TODO: Genericity
```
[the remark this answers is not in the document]
ScreamZ: Yeah, that's why I added TODO: Genericity, It's a reminder for myself, my bad sorry.

Genericity will allow the user to specicy something different than any if he which to during it's coding when he knows the expected resuklt.

kuzzle-sdk.d.ts
```diff
@@ -0,0 +1,111 @@
...
+        action: string;
+        body?: BodyObject
+        index?: string;
+        collection?: string;
+        _id?: string;
+        volatile?: object // TODO: volatile
+    }
+
+    interface RawKuzzleResponse {
+        _index: string,
```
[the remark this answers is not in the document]
ScreamZ: Allright, can you give me a documentation explaining the `RawKuzzleResponse` cited in the doc ? (as i asked in the PR comment with checkboxes ?)

kuzzle-sdk.d.ts
```diff
@@ -0,0 +1,111 @@
...
+        _type: string,
+        _id: string,
+        _version: number,
+        result: 'created' | "", // TODO:
```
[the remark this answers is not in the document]
ScreamZ: Any type ? Can you explain this deeply in the documentation ? Or here at least for me ? thanks

kuzzle-sdk.d.ts
ScreamZ: In that case it's up to the user to extends this definition type the following:

```ts
interface MyUpdatedConfig extends KuzzleOptions {
    myCustomFieldAdd: boolean;
}

File: kuzzle-sdk.d.ts (hunk 2 above)
Comment (ScreamZ) [reply kuzzle-sdk.d.ts]: I'll rename this, you're right but currently in reality this is only about the `queueable: true` parameter in each methods.

File: kuzzle-sdk.d.ts (hunk 1 above)
Comment (ScreamZ) [reply kuzzle-sdk.d.ts]: Otherwise I can also allow any other key that way it's a bit more flexible but less « secure ». Hum… in fact this solution is not blocking the first I gave, experimented user could still create an interface.

```ts
interface SquareConfig {
     host: string;
     autoQueue?: boolean;
     // …
    [propName: string]: any;
}
```

src/Kuzzle.js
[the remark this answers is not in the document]
Aschen: I think we can't have a proper type definition here because the options change for each protocol..

kuzzle-sdk.d.ts
Aschen: I'm ok with this solution :)

Raiper34:
Can you merge it please?

Aschen:
As @ScreamZ say, the version 6 is still in beta and may be subject to change.

But maybe we can still merge the typescript definitions now and update what has changed in another PR when we release the final version of the SDK. What do you think ?

ScreamZ:
@Aschen There is no point doing this atm, the file is too small and there are types errors which are result of API change.

I'll plan this update for half december

ScreamZ:
@Aschen @xbill82 @Njuelle @eric-trousset @scottinet @benoitvidis @ballinette This is updated.

kuzzle-sdk.d.ts
```diff
@@ -0,0 +1,292 @@
...
+    type RealtimeScope = "in" | "out"
+    interface SubscribeOptions {
+        scope?: "all" | RealtimeScope | "none";
+        users?: "all" | RealtimeScope | "none";
+        subscribeToSelf?: boolean;
+        volatile?: ObjectWithAnyKeys
+    }
+
+    interface KuzzleOptions {
+        host: string;
```
Aschen:
```suggestion
```

kuzzle-sdk.d.ts
```diff
@@ -0,0 +1,292 @@
...
+        requestId: string,
+        collection: string,
+        result: QueryResult
+        index: string,
+    }
+
+    interface WebsocketEventDocument extends RawKuzzleResponse {
+        timestamp: number,
+        volatile: ObjectWithAnyKeys,
+        protocol: "websocket" | string, // TODO: What else ?
```
Aschen:
```suggestion
         protocol: "websocket" | "mqtt",
```

kuzzle-sdk.d.ts
```diff
@@ -0,0 +1,292 @@
...
+        scope: RealtimeScope,
+        type: string, // TODO: Fix type enum
```
Aschen:
```suggestion
         type: "document" | "user"
```

kuzzle-sdk.d.ts
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        interface Ms extends ObjectWithAnyKeys { // TODO: Once ready
+
+        }
+
+        interface Realtime {
+            count(roomId: string, options?: QueryOptions): Promise<number>
+            publish(index: string, collection: string, message: ObjectWithAnyKeys, options?: QueryOptions): Promise<boolean>
+            subscribe(index: string, collection: string, filters: ObjectWithAnyKeys, callback: (eventObject: WebsocketEventDocument) => void, options?: SubscribeOptions): Promise<string>
```
Aschen:
```suggestion
             subscribe(index: string, collection: string, filters: ObjectWithAnyKeys, callback: (notification: Notification) => void, options?: SubscribeOptions): Promise<string>
```

kuzzle-sdk.d.ts
```diff
@@ -0,0 +1,292 @@
...
+        error: KuzzleError,
+        controller: string,
+        action: string,
+        requestId: string,
+        collection: string,
+        result: QueryResult
+        index: string,
+    }
+
+    interface WebsocketEventDocument extends RawKuzzleResponse {
```
Aschen:
```suggestion
     interface Notification extends RawKuzzleResponse {
```

ScreamZ:
@Aschen Can you give me documentation for Websocket constructor (and others protocols) ? I need it to make this clear, currently we have `port` in Kuzzle constructor options not sure this is good should be in protocol config.

ScreamZ:
Guys, this is almost completed, can you review ?

kuzzle-sdk.d.ts
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    export class Http implements KuzzleAbstractProtocol {
+        public readonly name: string;
+        constructor(host: string, options?: {
+            port?: number;
+            sslConnection?: boolean;
+        })
+    }
+    export class Websocket implements KuzzleAbstractProtocol {
```
Aschen:
```suggestion
    export class WebSocket implements KuzzleAbstractProtocol {
```

ScreamZ:
@scottinet This could be a solution : https://github.com/adamcarr/typescript-definition-tester#readme

scottinet:
This sounds promising indeed.

gpulido:
Hello,
After trying to use the defined d.ts I found that the KuzzleEvent enum definition is throwing an:
```cannot read property 'CONNECTED' of undefined```
After searching a bit I have found the following post explaining the situation:
https://lukasbehal.com/2017-05-22-enums-in-declaration-files/

Maybe this has to be changed before merging...
I would like to apologice if this is not a problem as I'm pretty new to typescript...

ScreamZ:
@gpulido I didn't had time to test it deeper, maybe .d.ts is not getting compiled but only used by linter.

I'll look forward once I have occasion to test it, Kuzzle team don't expect to merge it for now because they can't guarantee that typings will stay equivalent to the API, I'll probably release it on @types repository of microsoft somedays.

gpulido:
@ScreamZ I'm already testing it and it seems to work properly, I'm capable of importing types on Angular to use kuzzle. If I found any issue I will inform.
I guess, that we should be updating the d.ts as the api changes.
Thank you for the work!!

ScreamZ:
@gpulido We will have to do it on our own, i'll make a PR on definitelyTyped soon

PR #320, opened 2018-10-08
Fix return value for index#list and index#mDelete

Aschen:
## What does this PR do?

This PR fix the return value for index#list and index#mDelete.  
They now returns array of strings instead of an object containing only one property.

  - Step 1 : Create some indexes `for i in {1..10}; do; curl -H "Content-Type: application/json" -X POST "http://localhost:7512/test$i/_create" &; sleep 0.05; done`
  - Step 2 : List them `await kuzzle.index.list() ` and check return value
  - Step 3 : Delete them `await kuzzle.index.mDelete(['test1', 'test2']) and check return value

### DIFF
--- a/src/controllers/index.js
+++ b/src/controllers/index.js
@@ -70,7 +70,7 @@ class IndexController {
       controller: 'index',
       action: 'list'
     }, options)
-      .then(response => response.result);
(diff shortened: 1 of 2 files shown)


ScreamZ:
Nice one

PR #327, opened 2018-10-23
fix searchResult next

benoitvidis:
## What does this PR do?

This PR attempts to fix 2 small bugs in the `SearchResult.next` method

1. [from, size] If the from value is set to 0, the next won't return any further page
2. [sort, size] Only the top level entries of the sort options would be taken by the [`search_after`](https://www.elastic.co/guide/en/elasticsearch/reference/master/search-request-search-after.html) call

### Other changes

I had to update some dependencies for babel to make it build again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,11 +27,12 @@
   "main": "index.js",
   "license": "Apache-2.0",
   "dependencies": {
+    "lodash.get": "^4.4.2",
     "min-req-promise": "^1.0.1",
     "uws": "10.148.0"
   },
   "devDependencies": {
-    "babel-core": "^6.26.3",
+    "@babel/core": "^7.1.2",
     "babel-loader": "8.0.2",
     "babel-preset-env": "^1.7.0",
     "codecov": "^3.1.0",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
   "main": "index.js",
   "license": "Apache-2.0",
   "dependencies": {
-    "lodash.get": "^4.4.2",
     "min-req-promise": "^1.0.1",
     "uws": "10.148.0"
   },
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-sdk",
-  "version": "6.0.0-beta-2",
+  "version": "6.0.0-beta-3",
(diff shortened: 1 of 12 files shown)


src/controllers/searchResult/base.js
```diff
@@ -54,12 +55,14 @@ class SearchResultBase {
...
-        if (typeof sort === 'string') {
-          request.search_after.push(hit._source[sort]);
-        }
-        else {
-          request.search_after.push(hit._source[Object.keys(sort)[0]]);
-        }
+        const key = typeof sort === 'string'
+          ? sort
+          : Object.keys(sort)[0];
+        const value = key === '_id'
```
benoitvidis: _id is a valid sort field and it not part of the _source

package.json
```diff
@@ -27,11 +27,12 @@
   "main": "index.js",
   "license": "Apache-2.0",
   "dependencies": {
+    "lodash.get": "^4.4.2",
```
Aschen: This dependency add 4.5Kb to the bundle (https://bundlephobia.com/result?p=lodash.get@4.4.2).
Maybe you could just add a function like this one https://gist.github.com/Aschen/6ba04c9eca3051a83f7e1dbddf38c6a1

src/controllers/searchResult/base.js
```diff
@@ -54,12 +55,14 @@ class SearchResultBase {
...
-          request.search_after.push(hit._source[sort]);
-        }
-        else {
-          request.search_after.push(hit._source[Object.keys(sort)[0]]);
-        }
+        const key = typeof sort === 'string'
+          ? sort
+          : Object.keys(sort)[0];
+        const value = key === '_uid'
+          ? hit._id
```
ballinette: according to ES documentation, I understand that if results are sorted by `_uid`, the "search_after" item whould be `this._request.collection + '#" + hit._id` instead (see https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-request-search-after.html )
and we also should consider sorting by `_id`

benoitvidis: :+1:  good catch

package.json
benoitvidis: :+1:

PR #331, opened 2018-10-29
Change incoherent method returns

Aschen:
## What does this PR do?

Simplify returned values for the following methods:
 - `index:delete`
 - `index:setAutoRefresh`
 - `index:refreshInternal`
 - `index:refresh`
 - `auth:deleteMyCredentials`

### DIFF
--- a/test/controllers/auth.test.js
+++ b/test/controllers/auth.test.js
@@ -106,7 +106,7 @@ describe('Auth Controller', () => {
               action: 'deleteMyCredentials'
             }, options);
 
-          should(res).match({acknowledged: true});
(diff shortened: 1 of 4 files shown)


test/controllers/auth.test.js
```diff
@@ -106,7 +106,7 @@ describe('Auth Controller', () => {
               action: 'deleteMyCredentials'
             }, options);
 
-          should(res).match({acknowledged: true});
+            should(res).be.exactly(true);
```
Aschen:
```suggestion
          should(res).be.exactly(true);
```

PR #340, opened 2018-12-06
Move offline queue from protocol to kuzzle

benoitvidis:
## What does this PR do?

This PR refactors the offline queue mechanism of the Javascript sdk to move its logic from the protocols to Kuzzle.

### DIFF
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -81,6 +81,22 @@ class Kuzzle extends KuzzleEventEmitter {
     this.network.addListener('offlineQueuePop', data => this.emit('offlineQueuePop', data));
     this.network.addListener('queryError', (err, query) => this.emit('queryError', err, query));
 
+    // offline queue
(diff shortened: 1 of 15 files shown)


src/Kuzzle.js
```diff
@@ -347,38 +386,55 @@ class Kuzzle extends KuzzleEventEmitter {
...
+    let queuable = true;
+    if (options && options.queuable === false) {
+      queuable = false;
+    }
+
+    if (this.queueFilter) {
+      queuable = queuable && this.queueFilter(request);
+    }
+
+    if (this.queuing && queuable) {
```
scottinet: This means that if the SDK instance is queuing, but the request is not queuable, it is sent to the network.
I'm not sure that this is intended, I would rather think that the request is to be discarded.

And, if this is the case, then maybe the `discarded` event should be triggered.

benoitvidis: :+1:

PR #347, opened 2018-12-20
Re-order dependencies to correct emplacement.

ScreamZ:
## What does this PR do?

This is fixing dependencies that should be in `dev dependencies` as they are not related to runtime but build phase.

Regards


Aschen:
After talking with @scottinet, it appear that we have to keep these dependencies in order to build the SDK dist

Aschen:
I put the PR in wip so we don't forget to remove the dependencies later

scottinet:
@Aschen > we should write a new (internal) issue instead: once we release the SDK, we need a little more than just moving the dependencies back to the devDependencies section; we need also to properly configure the npm publish action, to force it to build the `dist` folder

scottinet:
@Aschen > we should write a new (internal) issue instead: once we release the SDK, we need a little more than just moving the dependencies back to the devDependencies section; we need also to properly configure the npm publish action, to force it to build the `dist` folder.

Aschen:
@scottinet Done: https://jira.kaliop.net/browse/KZL-983
I close this PR.

PR #355, opened 2019-01-08
React native compatibility

scottinet:
# Description

Fixes the following issues:

* `Symbol` is not supported in older versions of React Native: 
  * all usages of that keyword have been removed from the source code
  * force babel to consider all instances of `for...of` to be applied to arrays only, preventing it to use Symbol iterators
* babel was incorrectly configured in webpack, and so it was ignored. Now that babel is properly applied, the built library size increases from 55ko to 95ko, approx.
* create a dedicated babel configuration file, to properly separate webpack from babel


src/controllers/auth.js
```diff
@@ -1,7 +1,5 @@
 const User = require('./security/user');
 
-const _kuzzle = Symbol();
```
Aschen: What were the advantages of using symbols?

scottinet: To emulate a private property in javascript in a very, very fast way, instead of using Object.defineProperty, which is very slow.
Initially, that solution was applied to our internal Request object, because performances are paramount in that case. I don't really know why it was applied here.

I decided to get rid of any notion of privacy here: if anyone want to play with an undocumented, obviously internal value, well, it's their responsability.

webpack.config.js
```diff
@@ -1,52 +1,38 @@
...
 
 module.exports = {
+  name: 'browser',
   mode: 'production',
   entry: './index.js',
   output: {
     path: `${__dirname}/dist`,
     filename: 'kuzzle.js',
-    library: 'KuzzleSDK',
+    library: 'kuzzle-sdk',
```
xbill82: The value of the `library` attribute should be a valid Javascipt reference name (i.e. dashes will cause trouble). See https://webpack.js.org/configuration/output/#output-library

> The variable MyLibrary will be bound with the return value of your entry file, if the resulting output is included as a script tag in an HTML page.

webpack.config.js
```diff
@@ -1,52 +1,38 @@
...
     libraryTarget: 'umd'
```
xbill82: BTW, why this?

scottinet: I asked @Njuelle about this, as I don't know a thing about how webpacked modules work in browsers. :astonished:

webpack.config.js
xbill82:
```suggestion
    library: {
      root: 'KuzzleSDK',
      amd: 'kuzzle-sdk',
      commonjs: 'kuzzle-sdk'
    }
```

PR #359, opened 2019-01-14
Fixed call to checkToken after reconnection

xbill82:
Fixes a call to `checkToken` upon reconnection to Kuzzle. The previous version was surely a copy/paste error based on v5.

- connect to Kuzzle
- add a listener on the `reconnected` event
- disconnect from Kuzzle (one way to do it is to `docker-compose stop kuzzle`)
- reconnect to Kuzzle
- the callback attached to your listened should be triggered

### DIFF
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -275,7 +275,7 @@ class Kuzzle extends KuzzleEventEmitter {
       }
 
       if (this.jwt) {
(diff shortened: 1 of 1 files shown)


xbill82:
@scottinet yep, sorry I forgot to add the WIP tag... Just added the test!

Issue #366, opened 2019-01-27
[6.0.0-beta-6] kuzzle.eventActions no longer exists ?


lethak:
In prior versions, one could do 

```javascript
    const eventList = kuzzle.eventActions

    eventList.forEach((evtName) => {
        kuzzle.addListener(evtName, (arg1, arg2) => {
            console.log('[KUZZLE Event] ' + evtName + ' :: ', arg1, arg2)
        })
    })
```

this was thanks to:

https://github.com/kuzzleio/sdk-javascript/blob/7ac7c1c300be2fa4b18c19c24de1266e5d7c7bae/src/Kuzzle.js#L55

but no longer possible with:

https://github.com/kuzzleio/sdk-javascript/blob/d1f54e26b17151875ba671134348d3f908033dbd/src/Kuzzle.js#L16

May I suggest keeping them public and read only ?
thanks

Aschen:
Hi @lethak !
Yes, this is totally something we can do. Thanks for the feedback :)

Aschen:
https://github.com/kuzzleio/sdk-javascript/pull/367

Aschen:
Won't do for now.  
If someone else need this feature, please comment this issue.

PR #367, opened 2019-01-28
Expose events list

Aschen:
## What does this PR do?

Expose the list of SDK events in the `eventList` property.

See this issue https://github.com/kuzzleio/sdk-javascript/issues/366
Doc: https://github.com/kuzzleio/documentation-V2/pull/224

### DIFF
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -12,20 +12,6 @@ const
   BaseController = require('./controllers/base'),
   uuidv4 = require('./uuidv4');
 
-const
-  events = [
-    'connected',
-    'discarded',
-    'disconnected',
-    'loginAttempt',
-    'networkError',
-    'offlineQueuePush',
-    'offlineQueuePop',
-    'queryError',
-    'reconnected',
-    'tokenExpired'
-  ];
-
 class Kuzzle extends KuzzleEventEmitter {
 
   /**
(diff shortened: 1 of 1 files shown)


src/Kuzzle.js
```diff
@@ -47,6 +33,19 @@ class Kuzzle extends KuzzleEventEmitter {
     }
     this.protocol = protocol;
 
+    this._eventList = [
```
lethak: I think you could have kept it in the node module scope with

```javascript
const		
   events = [		
     'connected',		
     'discarded',		
     'disconnected',		
     'loginAttempt',		
     'networkError',		
     'offlineQueuePush',		
     'offlineQueuePop',		
     'queryError',		
     'reconnected',		
     'tokenExpired'		
   ];
```

and simply implement a getter like 

```javascript
 get eventList () {
     return events
   }
```

without the need to modify anything else.

I know the use case of having multiple instance of the Kuzzle object is low, but this way, the memory fingerprint could be lighter since the array is not implemented for each instance.

Aschen: Yes you're right !
I did it like that because it's more logical to have the event list inside the kuzzle instance and not in the module global scope. In the futur we might want to let developer use middleware with the SDK and these middleware could add new events for example.

Aschen:
For the other SDK we can just expose the same getter, I didn't see the problem?
Concerning the usecase, maybe you can be more explicit @lethak ?

Aschen:
It's useful to debug events in development mode.  
I think we can have slightly different api across SDKs, specially with the JS one who is a little particular since it does not derive from SWIG.

scottinet:
I don't get your argument about debugging events: who will debug events? SDK users?
Because I don't think it's their role to debug the SDK: users should expect the SDK to work properly, and correctly trigger events.

To me, there is still no real incentive to change our API. I don't have any other NPM module in mind that does that.

PR #369, opened 2019-02-07
Addlistener after connect

Aschen:
## What does this PR do?

When someone instantiate Kuzzle, we start to add listener on events.  
With SocketIO protocol, these events must be added also on the `socket` but the `socket` itself is instantiated only after a call to `connect`

  - Step 1 : Instantiate Kuzzle with SocketIO protocol


test/kuzzle/constructor.test.js
```diff
@@ -134,4 +139,22 @@ describe('Kuzzle constructor', () => {
     should(kuzzle.autoQueue).be.true();
     should(kuzzle.autoReplay).be.true();
   });
+
+  it('should initialize kuzzle with SocketIO', () => {
+    new Kuzzle(
+      new SocketIO('localhost')
```
scottinet: what does this test test?

Aschen: Just test that we can instantiate Kuzzle with SocketIO.  
This PR fix an error on Kuzzle instantiation with SocketIO

scottinet: I'm reluctant validating a test that does not include `should` checks. Neither the test title nor its content make it clear what is tested. Especially, I see no correlation between the events being registered at a later time, and the instantiation of a protocol subclass.

Aschen: I just added these tests because from the beginning the SocketIO protocol has never work because of this bug when instantiating the SDK with this protocol.  
I removed it

scottinet: What I meant is that any maintainer in the future would probably remove these tests out of clues of what these tests test.
Removing those is one way to "fix" the issue, but you could also have left a comment, or used a `should` test such as `should(...).not.throw()` or `should(...).instanceOf(...)`.

test/kuzzle/constructor.test.js
```diff
@@ -139,22 +139,4 @@ describe('Kuzzle constructor', () => {
     should(kuzzle.autoQueue).be.true();
     should(kuzzle.autoReplay).be.true();
   });
-
-  it('should initialize kuzzle with SocketIO', () => {
```
alexandrebouthinon: Why did you remove those tests?

test/kuzzle/constructor.test.js
Aschen: Got it :)

PR #370, opened 2019-02-08
[KZL-996] Save Kuzzle API routes in a json file

alexandrebouthinon:
## What does this PR do?
Save Kuzzle API routes in a json file to permit the HTTP protocol to not have to performs request on Kuzzle to fetch all available `vanilla` routes.

Run Unit test and/or use HTTP protocol

### DIFF
--- a/src/protocols/http.js
+++ b/src/protocols/http.js
@@ -1,45 +1,8 @@
 'use strict';
 
 const
-  KuzzleAbstractProtocol = require('./abstract/common');
-
-const
-  _routes = {
-    routes: {
-      auth: {
-        login: {
-          verb: 'POST',
-          url: '/_login/:strategy'
-        }
-      },
-      bulk: {
-        import: {
-          verb: 'POST',
-          url: '/:index/:collection/_bulk'
-        }
-      },
-      document: {
-        create: {
-          verb: 'POST',
-          url: '/:index/:collection/_create'
-        }
-      },
-      security: {
-        createFirstAdmin: {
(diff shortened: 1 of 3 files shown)


src/protocols/http.js
```diff
@@ -78,25 +41,7 @@ class HttpWrapper extends KuzzleAbstractProtocol {
     }
 
     return this._sendHttpRequest('GET', '/')
```
Aschen: This is not used anymore, you should remove this call because it may throw an error if the anonymous user does not have the server:info right

alexandrebouthinon:
I put this PR on WIP. After quick talk with @Aschen, it sounds cool to have a new route in Kuzzle server controller to return plugins customs routes.

scottinet:
The issue resolved by this PR here is that, without rights to `server:info`, the HTTP part of the SDK is unable to perform correctly.
Creating a dedicating API route allowing kuzzle to send the available routes just moves the problem, it doesn't solve it.

Moreover, between `server:info` and the swagger routes, we should already be covered.

alexandrebouthinon:
@Aschen : Does the @scottinet's answer satisfied you?

src/protocols/http.js
alexandrebouthinon: Anonymous can use the HTTP protocol since this function will not reject in case of permissions issue. You'll still get a response from the server (with 403 status code yes but a response). This will rejects only in case of network issue. :slightly_smiling_face:

PR #372, opened 2019-02-13
Don't send jwt when doing the login action

jenow:
## What does this PR do?

Don't send the jwt when doing a `login` action. This prevent to be able to do a login after a `tokenExpired` happen without resetting the jwt (kuzzle.jwt = null)


scottinet:
While ignoring the JWT on a login action is not a bad idea, I think there is another problem with the use case you mention.

Namely, the JWT should be unset after a `tokenExpired` event has been received: https://github.com/kuzzleio/sdk-javascript/blob/6-beta/src/Kuzzle.js#L93

So, if you're right about your use case, then it means that there is a bug with that event.

scottinet:
Ignore my previous comment: I'll fix that issue in another PR. After discussing the issue internally, not sending the JWT in login requests seems a reasonable thing to do.

PR #373, opened 2019-02-14
WebSocket: fix error object on connection error

scottinet:
# Description

On a websocket connection error, the error thrown by the SDK has the following message: `[object Object]`

This is because on a connection error, `ws` sends a complex object, but our error handler processes it as if it's a string.

# How to test it

Run this snippet:

```js
const {Kuzzle, WebSocket} = require('kuzzle-sdk');
const
  ws = new WebSocket('localhost'),
  kuzzle = new Kuzzle(ws);

kuzzle.connect().catch(err => console.error(err));
kuzzle.disconnect();
```

Before this PR, this prints: 

```
Error: [object Object]
    ... stack trace ...
```

With this PR:

```
Error: WebSocket was closed before the connection was established
    ... stack trace ...
```

### DIFF
--- a/src/protocols/websocket.js
+++ b/src/protocols/websocket.js
@@ -86,7 +86,8 @@ class WSNode extends RTWrapper {
       };
 
       this.client.onerror = error => {
-        const err = (error instanceof Error) && error || new Error(error);
+        const err = error ?
+          new Error(error.message || error) : new Error('Unexpected error');
 
         this.clientNetworkError(err);
 
--- a/src/protocols/websocket.js
+++ b/src/protocols/websocket.js
@@ -86,8 +86,12 @@ class WSNode extends RTWrapper {
       };
 
       this.client.onerror = error => {
-        const err = error ?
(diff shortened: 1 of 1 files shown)


src/protocols/websocket.js
```diff
@@ -86,7 +86,8 @@ class WSNode extends RTWrapper {
       };
 
       this.client.onerror = error => {
-        const err = (error instanceof Error) && error || new Error(error);
+        const err = error ?
+          new Error(error.message || error) : new Error('Unexpected error');
```
Aschen: Don't we loose the stacktrace returned by Kuzzle?

Aschen: It will be fixed with https://jira.kaliop.net/browse/KZL-1001, nevermind

scottinet: Well, even if it will be fixed, I think you made a valid point. I prefer to fix this now, even if it is redundant with a future task.

scottinet: @Aschen > fixed

PR #384, opened 2019-04-08
[fix] search API: "sort" and "search_after" must be in the requests body

scottinet:
# Description

The `SearchResult` class sends a `search_after` request to Kuzzle upon calling its `next` method, if:
* no scroll_id is available
* a `size` option is provided
* the search query contains a `sort` modifier

The current implementation considers that `sort` must be at the root of the API request, while Kuzzle expects it inside the body part of it.
Same thing goes for the `search_after` argument, which must be put in the request body, not at root level.

### DIFF
--- a/src/controllers/searchResult/base.js
+++ b/src/controllers/searchResult/base.js
@@ -45,15 +45,16 @@ class SearchResultBase {
           return this;
         });
     }
-    else if (this._request.size && this._request.sort) {
+    else if (this._request.size && this._request.body.sort) {
       const
         request = Object.assign({}, this._request, {
-          action: this._searchAction,
-          search_after: []
+          action: this._searchAction
         }),
-        hit = this._response.hits && this._response.hits[this._response.hits.length -1];
+        hit = this._response.hits[this._response.hits.length -1];
 
-      for (const sort of this._request.sort) {
+      request.body.search_after = [];
+
+      for (const sort of this._request.body.sort) {
         const key = typeof sort === 'string'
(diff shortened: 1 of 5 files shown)


src/controllers/searchResult/base.js
```diff
@@ -45,23 +45,24 @@ class SearchResultBase {
...
+        hit = this._response.hits[this._response.hits.length -1];
 
-      for (const sort of this._request.sort) {
+      request.body.search_after = [];
+
+      for (const sort of this._request.body.sort) {
         const key = typeof sort === 'string'
           ? sort
           : Object.keys(sort)[0];
         const value = key === '_uid'
```
Yoann-Abbes: (nitpicking) maybe adding parentheses make it more readable
```suggestion
        const value = (key === '_uid')
```

src/controllers/searchResult/base.js
```diff
@@ -45,23 +45,24 @@ class SearchResultBase {
...
+          action: this._searchAction
         }),
-        hit = this._response.hits && this._response.hits[this._response.hits.length -1];
+        hit = this._response.hits[this._response.hits.length -1];
 
-      for (const sort of this._request.sort) {
+      request.body.search_after = [];
+
+      for (const sort of this._request.body.sort) {
         const key = typeof sort === 'string'
```
Yoann-Abbes: maybe adding parentheses here make it more readable

```suggestion
        const key = (typeof sort === 'string')
```

PR #386, opened 2019-04-16
Add SearchResult.forEachHit

Aschen:
## What does this PR do?

Adds a `SearchResult.forEachHit` method that take a callback action and execute it over every hit of the search.  
This method iterate over the page of results with the `next()` method and call the provided action of every elements of the `results.hits` array.

When we use the `document:search` method, a very current pattern is to apply an action on every results of the search whatever the number of pages. 

```
while (results) {
  for (const hit of results.hits) {
    /* do something with hit */
  }
  results = await results.next()
}
```

Doc: https://github.com/kuzzleio/documentation/pull/281

  - Step 1 : Do some paginated search
  - Step 2 : use `results.forEachHit(hit => console.log(hit._id))`
  - Step 3 :


src/controllers/searchResult/base.js
```diff
@@ -97,6 +95,35 @@ class SearchResultBase {
...
+    while (results) {
+      for (const hit of results.hits) {
+        const ret = action(hit);
+
+        if (ret && typeof ret.then === 'function') {
+          promises.push(ret);
+        }
+      }
+
+      results = await results.next();
```
benoitvidis: Don't we support node 6?

Aschen: Node 6 LTS support ends on the 30/04/2019, but I will use simple promise instead since this is the only `async/await` of the SDK

src/controllers/searchResult/base.js
```diff
@@ -97,6 +95,46 @@ class SearchResultBase {
...
+   * be resolved.
+   *
+   * @param {Function} action - Action to execute for each hit
+   * @returns {Promise}
+   */
+  async forEachHit(action, firstCall = true) {
+    let results;
+
+    if (firstCall) {
+      results = new this.constructor(this._kuzzle, this._request, this._options, this._response);
```
ballinette: why do we need to instantiate a new `SearchResult` object here ?

Aschen: The `next` method does not return a new `SearchResult` but instead it modify the properties.  
I instantiate a new `SearchResult` to avoid mutating the original one when calling `forEachHit`

scottinet: I agree, the `next` method should return a new `SearchResult` instance, as it is done in other SDK, instead of mutating the current object.

scottinet:
I can't help but think that all your feature does, is to get rid of one perfectly concise and clear `do...while` loop. And I find that underwhelming.

Going down that route, why not also add `mapHits`, `filterHits`, and other methods from the Array prototype?

Anyway, we'll need the team to review that one, because I don't think we can solve this disagreement by ourselves.

PR #388, opened 2019-04-23
Use BaseController class for controllers

Aschen:
## What does this PR do?

*I'm not sure if this is a new feature or an enhancement* 

This PR intend to unify the way we add/use controllers with the SDK. In this way we can have the same API when using our core controllers (`document`, `collection`, etc.) and custom controllers made by users.  

All existing controllers are now inheriting from the `BaseController` class and they are added to the SDK with `Kuzzle.useController` instead of been manually instantiated.

Example of custom controller
```
class TaxiController extends BaseController {
  constructor (kuzzle) {
    super(kuzzle, 'my-plugin/taxi');
  }

  startDuty (driver) {
    return this.query({ driver, action: 'startDuty' });
  }
}

kuzzle.useController(TaxiController, 'taxi');

kuzzle.taxi.startDuty('Martin');
```

I know this look like a breaking change but since we never documented this feature by now it is not really one.

Documentation https://github.com/kuzzleio/documentation/pull/284

  - Step 1 : Run unit tests

### DIFF
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -62,15 +62,15 @@ class Kuzzle extends KuzzleEventEmitter {
     this.volatile = typeof options.volatile === 'object' ? options.volatile : {};
 
     // controllers
-    this.auth = new AuthController(this);
-    this.bulk = new BulkController(this);
-    this.collection = new CollectionController(this);
-    this.document = new DocumentController(this);
-    this.index = new IndexController(this);
-    this.ms = new MemoryStorageController(this);
-    this.realtime = new RealtimeController(this);
-    this.security = new SecurityController(this);
-    this.server = new ServerController(this);
+    this.useController(AuthController, 'auth');
+    this.useController(BulkController, 'bulk');
+    this.useController(CollectionController, 'collection');
+    this.useController(DocumentController, 'document');
+    this.useController(IndexController, 'index');
+    this.useController(MemoryStorageController, 'ms');
+    this.useController(RealtimeController, 'realtime');
+    this.useController(SecurityController, 'security');
+    this.useController(ServerController, 'server');
 
     // offline queue
     this._autoQueue = typeof options.autoQueue === 'boolean' ? options.autoQueue : false;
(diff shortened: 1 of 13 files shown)


src/Kuzzle.js
```diff
@@ -443,28 +443,34 @@ Discarded request: ${JSON.stringify(request)}`));
...
-  useController (controller) {
-    if (!(controller instanceof BaseController)) {
+  useController (ControllerClass, accessor) {
+    if (!(ControllerClass.prototype instanceof BaseController)) {
       throw new Error('Controllers must inherits from the BaseController class.');
     }
 
-    if (!(controller.name && controller.name.length > 0)) {
-      throw new Error('Controllers must have a name.');
+    if (!(accessor && accessor.length > 0)) {
```
Yoann-Abbes: Didn't you mean
```suggestion
    if (!(accessor && accessor.length <= 0)) {
```

?

src/Kuzzle.js
```diff
@@ -443,28 +443,34 @@ Discarded request: ${JSON.stringify(request)}`));
...
+
+    if (this[accessor]) {
+      throw new Error(`There is already a controller with the accessor '${accessor}'. Please use another one.`);
     }
 
-    if (!(controller.accessor && controller.accessor.length > 0)) {
-      throw new Error('Controllers must have an accessor.');
+    const controller = new ControllerClass(this);
+
+    if (!(controller.name && controller.name.length > 0)) {
```
Yoann-Abbes: Didn't you mean
```suggestion
    if (!(controller.name && controller.name.length <= 0)) {
```
?

src/Kuzzle.js
Aschen: No ^^ Don't you see the `!` ?

src/Kuzzle.js
Aschen: No ^^

src/Kuzzle.js
Yoann-Abbes: :O my bad...

src/Kuzzle.js
Yoann-Abbes: I was just testing you !

Aschen:
@scottinet I didn't understand what you mean by explicitly declare what actions can be performed for that registering controller. Since the controller is a class, is not the SDK responsibility to allow or disallow method calling. If we weren't in JS, we would have the `private` keyword but..

I could wrap the controller class with a Proxy or something but I didn't really see the point here.

scottinet:
@Aschen > You have a point. 
Just to explain myself: we had to make actions exposure explicit in Kuzzle's controllers to prevent internal methods from being leaked to the API. And what I erroneously thought that we had possibly the same problem, while it's obvious that the context is different.

PR #389, opened 2019-04-26
Remove usage of _meta

Aschen:
## What does this PR do?

Remove the usage of _meta property.

This is not a breaking change since this property wasn't documented


scottinet:
@stafyniaksacha > the meta are still there, in `User.content._kuzzle_info`, it's just the undocumented and redundant `User.meta` property that is being removed (I had the same reaction than you initially).

PR #390, opened 2019-04-26
Add authenticated property on Kuzzle object

Aschen:
## What does this PR do?

 - Add `authenticated` property on the Kuzzle object
 - Refactor the Auth controller to keep responsibility for the authentication token

  - Step 1 :
  - Step 2 :
  - Step 3 :

### DIFF
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -10,7 +10,8 @@ const
   SecurityController = require('./controllers/security'),
   MemoryStorageController = require('./controllers/memoryStorage'),
   BaseController = require('./controllers/base'),
-  uuidv4 = require('./uuidv4');
+  uuidv4 = require('./uuidv4'),
+  proxify = require('./proxify');
 
 const
   events = [
@@ -87,6 +88,14 @@ class Kuzzle extends KuzzleEventEmitter {
     }
     this.queuing = false;
     this.replayInterval = 10;
+
+    this._jwt = undefined;
+
+    return proxify(this, {
+      seal: true,
+      name: 'kuzzle',
+      exposeApi: true
+    });
   }
 
   get autoQueue () {
@@ -446,20 +455,23 @@ Discarded request: ${JSON.stringify(request)}`));
       throw new Error('You must provide a valid accessor.');
     }
 
-    if (this[accessor]) {
+    if (this.__proxy__ ? this.__proxy__.hasProp(accessor) : this[accessor]) {
       throw new Error(`There is already a controller with the accessor '${accessor}'. Please use another one.`);
     }
 
     const controller = new ControllerClass(this);
-
+    
     if (!(controller.name && controller.name.length > 0)) {
       throw new Error('Controllers must have a name.');
     }
-
+    
     if (controller.kuzzle !== this) {
       throw new Error('You must pass the Kuzzle SDK instance to the parent constructor.');
     }
-
+    
+    if (this.__proxy__) {
+      this.__proxy__.registerProp(accessor);
+    }
     this[accessor] = controller;
 
     return this;
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -284,11 +284,11 @@ class Kuzzle extends KuzzleEventEmitter {
             // shouldn't obtain an error but let's invalidate the token anyway
             if (!res.valid) {
               this.jwt = undefined;
-                    }
+            }
           })
           .catch(() => {
             this.jwt = undefined;
-                })
+          })
           .then(() => this.emit('reconnected'));
       }
 
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -203,6 +203,15 @@ class Kuzzle extends KuzzleEventEmitter {
     return this.protocol.sslConnection;
   }
 
+  isLoggued () {
+    if (!this.jwt) {
+      return false;
+    }
+
+    return this.auth.checkToken(this.jwt)
+      .then(({ valid }) => valid);
+  }
+
   /**
   * Emit an event to all registered listeners
   * An event cannot be emitted multiple times before a timeout has been reached.
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -208,8 +208,7 @@ class Kuzzle extends KuzzleEventEmitter {
       return false;
     }
 
-    return this.auth.checkToken(this.jwt)
-      .then(({ valid }) => valid);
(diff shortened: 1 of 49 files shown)


src/Kuzzle.js
```diff
@@ -239,6 +247,7 @@ class Kuzzle extends KuzzleEventEmitter {
 
     this.protocol.addListener('tokenExpired', () => {
       this.jwt = undefined;
+      this.jwtExpiresAt = undefined;
```
alexandrebouthinon: It's just question from a JS noob, what is the difference between:
```js
this.jwtExpiresAt = undefined;
```
and
```js
delete this.jwtExpiresAt;
```

Aschen: With `this.jwt = undefined` the function `hasOwnProperty('jwt')` will still return true but in this case it's change nothing

src/Kuzzle.js
```diff
@@ -203,6 +203,14 @@ class Kuzzle extends KuzzleEventEmitter {
     return this.protocol.sslConnection;
   }
 
+  isLoggued () {
+    if (!this.jwt) {
+      return false;
+    }
+
+    return this.jwtExpiresAt > Date.now();
```
scottinet: This falls short as soon as the `jwt` property is set with a custom token (i.e. saved from an earlier connection), which is a very, very common situation.

For this to work, you must also update the `jwt` setter, decode the token's payload and retrieve its timeout value.

src/Kuzzle.js
```diff
@@ -203,6 +203,10 @@ class Kuzzle extends KuzzleEventEmitter {
     return this.protocol.sslConnection;
   }
 
+  get authenticated () {
+    return this.jwt !== null && this.jwt !== undefined;
```
scottinet: (nitpicking) according to the jwt setter code, that property can never be null

src/Kuzzle.js
```diff
@@ -203,6 +181,22 @@ class Kuzzle extends KuzzleEventEmitter {
     return this.protocol.sslConnection;
   }
 
+  get authenticated () {
+    return this.auth.authenticationToken && !this.auth.authenticationToken.expired;
+  }
+
+  get jwt () {
+    if (!this.auth.authenticationToken) {
```
scottinet: (nitpicking) should always be null if the token is not set, so this test is redundant

src/controllers/auth.js
```diff
@@ -24,11 +58,15 @@ class AuthController extends BaseController {
    * @param  {string}   token     The jwt token to check
    * @return {Promise|*|PromiseLike<T>|Promise<T>}
    */
-  checkToken (token) {
+  checkToken (token = undefined) {
```
scottinet: The default parameter is useless: it'll set token to undefined only if it is... undefined.

src/controllers/auth.js
```diff
@@ -24,11 +58,15 @@ class AuthController extends BaseController {
...
+    if (token === undefined && this.authenticationToken) {
```
scottinet: (to be discussed) Personnally, I would throw if no token is defined or, perhaps better, I would let Kuzzle respond that it needs a token to be verified (like it was done before your changes)

And if users want to verify the stored token, they only need to use it like this: `kuzzle.auth.checkToken(kuzzle.jwt)`
Don't forget that the SDK automatically does an auth:checkToken on the stored jwt when a network connection is (re-)established.

src/core/Jwt.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  get expired () {
+    return Date.now() > this.expiresAt;
+  }
+
+  _decode () {
+    const 
+      [, payloadRaw, ] = this._encodedJwt.split('.'),
+      payload = JSON.parse(new Buffer(payloadRaw, 'base64').toString());
```
scottinet: you first need to check that `payloadRaw` is not undefined (invalid token) and you also need to add a `try..catch` around JSON.parse, in case the payload is not in JSON format (again, invalid token).

test/mocks/generateJwt.mock.js
```diff
@@ -0,0 +1,14 @@
+function generateJwt (userId = 'test-user', expiresAt = null) {
+  const 
+    header = { alg: 'HS256', typ: 'JWT' },
+    payload = { _id: userId, iat: Date.now(), exp: expiresAt },
+    signature = 'who care ?';
```
scottinet: (nitpicking)

```suggestion
    signature = 'who cares?';
```

src/Kuzzle.js
Aschen: But if I didn't check if `authenticationToken` is set, I will have TypeError when I try to access `encodedJwt`

src/controllers/auth.js
Aschen: I know this was the original behaviour but most of the time users want to check if the token stored by the SDK is still valid so IMHO they will understand that if you call this method without a specific token, it will check the SDK internal token.
Also it's more simpler to get ride of the `kuzzle.jwt` argument.
But maybe I can throw an error if `checkToken` is called without a token and if there is no internal token

src/Kuzzle.js
scottinet: Oh, yes, I missed the fact that you now stores the entire token object and not only the encoded value :+1:

src/controllers/auth.js
scottinet: The token is already automatically checked by the SDK, but... why not In that case, you have to document it properly in the checkToken SDK documentation

src/core/Jwt.js
```diff
@@ -0,0 +1,48 @@
...
+  _decode () {
+    const  [, payloadRaw, ] = this._encodedJwt.split('.');
+
+    if (!payloadRaw) {
+      throw new Error('Invalid JWT format');
+    }
+    
+    let payload;
+    try {
+      payload = JSON.parse(new Buffer(payloadRaw, 'base64').toString());
```
scottinet: I missed that one, but `new Buffer` is deprecated, you should use Buffer.from instead:

```suggestion
      payload = JSON.parse(Buffer.from(payloadRaw, 'base64').toString());
```

Also, Buffer does not exist on browsers, does Babel correctly handle it?
If not, you may want to check if `Buffer` exists in the global scope and, if it doesn't, use btoa instead

Aschen:
@Yoann-Abbes and @alexandrebouthinon You might want to review this PR again since I extracted the JWT from Kuzzle object

PR #395, opened 2019-05-17
[KZL-1098] proxify kuzzle to avoid mistyping error

thomasarbona:
#### `proxify` helper

Create a helper that can throw error when get/set undeclared properties.

```javascript
const obj = proxify({ foo: 42 });

console.log(obj.foo);
console.log(obj.bar); // throw
```

Some configuration:

```javascript
proxify({ foo: 42 }, {
  name: 'object',
  seal: true,
  sealGet: false,
  deprecated: [],
  exposeApi: false,
  apiNamespace: '__proxy__',
});
```

- `name` : the name of the proxified object for warnings
- `seal`: does `proxify` throw error on set undeclared properties?
- `sealGet`: does `proxify` throw error on get undeclared properties? this options is separated from `seal` because getting undefined properties can be usefull for type checking for example
- `deprecated`: array of property names which produce a deprecate warning on get/set
- `warnDepreciationOnce`: only warn once per deprecated property
- `exposeApi`: expose an api in the object to manipulate properties (described below)
- `apiNamespace`: in which namespace api is exposed

##### proxyfy API

```javascript
const obj = proxify({ foo: 42 }, { exposeApi: true });

console.log(obj.foo); // OK
console.log(obj.bar); // throw error

obj.__proxy__.registerProp('bar');

console.log(obj.bar); // OK

const hasProp = obj.__proxy__.hasProp('baz'); // check WITHOUT warning

obj.__proxy__.unregisterProp('bar');

console.log(obj.bar); // throw error
```

#### usage on kuzzle instance
kuzzle instance use `proxify` like this in the constructor:
```javascript
return proxify(this, {
  seal: true,
  name: 'kuzzle',
  exposeApi: true
});
```

  - Step 1 : instantiate `Kuzzle` class
  - Step 2 : define an undeclared property on this instance, like `foo`

### DIFF
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -10,7 +10,8 @@ const
   SecurityController = require('./controllers/security'),
   MemoryStorageController = require('./controllers/memoryStorage'),
   BaseController = require('./controllers/base'),
-  uuidv4 = require('./uuidv4');
+  uuidv4 = require('./uuidv4'),
+  proxify = require('./proxify');
 
 const
   events = [
@@ -87,6 +88,14 @@ class Kuzzle extends KuzzleEventEmitter {
     }
     this.queuing = false;
     this.replayInterval = 10;
+
+    this._jwt = undefined;
+
+    return proxify(this, {
+      seal: true,
+      name: 'kuzzle',
+      exposeApi: true
+    });
   }
 
   get autoQueue () {
@@ -460,20 +469,23 @@ Discarded request: ${JSON.stringify(request)}`));
       throw new Error('You must provide a valid accessor.');
     }
 
-    if (this[accessor]) {
+    if (this.__proxy__ ? this.__proxy__.hasProp(accessor) : this[accessor]) {
       throw new Error(`There is already a controller with the accessor '${accessor}'. Please use another one.`);
     }
 
     const controller = new ControllerClass(this);
-
+    
     if (!(controller.name && controller.name.length > 0)) {
       throw new Error('Controllers must have a name.');
     }
-
+    
     if (controller.kuzzle !== this) {
       throw new Error('You must pass the Kuzzle SDK instance to the parent constructor.');
     }
-
+    
+    if (this.__proxy__) {
+      this.__proxy__.registerProps(accessor);
+    }
     this[accessor] = controller;
 
     return this;
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -203,6 +203,10 @@ class Kuzzle extends KuzzleEventEmitter {
     return this.protocol.sslConnection;
   }
 
+  get connected () {
+    return this.protocol.connected;
+  }
+
   /**
   * Emit an event to all registered listeners
   * An event cannot be emitted multiple times before a timeout has been reached.
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -62,15 +62,15 @@ class Kuzzle extends KuzzleEventEmitter {
     this.volatile = typeof options.volatile === 'object' ? options.volatile : {};
 
     // controllers
-    this.auth = new AuthController(this);
-    this.bulk = new BulkController(this);
-    this.collection = new CollectionController(this);
-    this.document = new DocumentController(this);
-    this.index = new IndexController(this);
-    this.ms = new MemoryStorageController(this);
-    this.realtime = new RealtimeController(this);
-    this.security = new SecurityController(this);
-    this.server = new ServerController(this);
+    this.useController(AuthController, 'auth');
+    this.useController(BulkController, 'bulk');
+    this.useController(CollectionController, 'collection');
+    this.useController(DocumentController, 'document');
+    this.useController(IndexController, 'index');
+    this.useController(MemoryStorageController, 'ms');
+    this.useController(RealtimeController, 'realtime');
+    this.useController(SecurityController, 'security');
+    this.useController(ServerController, 'server');
 
     // offline queue
     this._autoQueue = typeof options.autoQueue === 'boolean' ? options.autoQueue : false;
@@ -443,28 +443,34 @@ Discarded request: ${JSON.stringify(request)}`));
   /**
    * Adds a new controller and make it available in the SDK.
    *
-   * @param {BaseController} controller
+   * @param {BaseController} ControllerClass
+   * @param {string} accessor
    * @returns {Kuzzle}
    */
-  useController (controller) {
-    if (!(controller instanceof BaseController)) {
+  useController (ControllerClass, accessor) {
+    if (!(ControllerClass.prototype instanceof BaseController)) {
       throw new Error('Controllers must inherits from the BaseController class.');
     }
+
+    if (!(accessor && accessor.length > 0)) {
+      throw new Error('You must provide a valid accessor.');
+    }
+
+    if (this[accessor]) {
+      throw new Error(`There is already a controller with the accessor '${accessor}'. Please use another one.`);
+    }
+
+    const controller = new ControllerClass(this);
 
     if (!(controller.name && controller.name.length > 0)) {
       throw new Error('Controllers must have a name.');
     }
 
-    if (!(controller.accessor && controller.accessor.length > 0)) {
-      throw new Error('Controllers must have an accessor.');
+    if (controller.kuzzle !== this) {
+      throw new Error('You must pass the Kuzzle SDK instance to the parent constructor.');
     }
 
-    if (this[controller.accessor]) {
-      throw new Error(`There is already a controller with the accessor '${controller.accessor}'. Please use another one.`);
-    }
-
-    controller.kuzzle = this;
-    this[controller.accessor] = controller;
+    this[accessor] = controller;
 
     return this;
   }
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -484,7 +484,7 @@ Discarded request: ${JSON.stringify(request)}`));
     }
     
     if (this.__proxy__) {
-      this.__proxy__.registerProps(accessor);
+      this.__proxy__.registerProp(accessor);
     }
     this[accessor] = controller;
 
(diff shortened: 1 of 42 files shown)


src/proxify.js
```diff
@@ -0,0 +1,81 @@
...
+    ...getPropertyNames(obj)
+  ]));
+
+  const handler = {
+    get: (target, name) => {
+      if (options.sealGet && typeof name === 'string' && !properties.includes(name)) {
+        throw new Error(`${options.name}.${name} is not defined`);
+      }
+      if (options.deprecated.includes(name)) {
+        console.warn(`Warning: ${options.name}.${name} is deprecated`);
```
scottinet: you must add some kind of flag to trigger that warning only once per deprecated property accessed, otherwise this can clog the console up pretty quickly

src/proxify.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+  
+  properties.push.apply(properties, deleteDuplicates([
+    ...Object.getOwnPropertyNames(obj),
+    ...getPropertyNames(obj)
+  ]));
+
+  const handler = {
+    get: (target, name) => {
+      if (options.sealGet && typeof name === 'string' && !properties.includes(name)) {
```
scottinet: you should remove the typecheck made on `name`: 
* it can only be a string or a symbol anyway, since this is a getter trap
* if I pass a non-existing symbol to your handler, I won't get any error even with `sealGet` active, which does not seem to be the intended behavior

src/proxify.js
```diff
@@ -0,0 +1,81 @@
...
+        console.warn(`Warning: ${options.name}.${name} is deprecated`);
+      }
+      return target[name];
+    },
+    set: (target, name, value) => {
+      if (options.seal && !properties.includes(name)) {
+        throw new Error(`setting a not defined '${name}' properties in '${options.name}' object`)
+      }
+      if (options.deprecated.includes(name)) {
+        console.warn(`Warning: ${options.name}.${name} is deprecated`);
```
scottinet: ditto

test/kuzzle/proxy.test.js
```diff
@@ -0,0 +1,31 @@
...
+    SocketIO,
+    WebSocket,
+    Http
+  } = require('../../src/protocols'),
+  ProtocolMock = require('../mocks/protocol.mock');
+
+describe('Kuzzle proxy', () => {
+  const protocolMock = new ProtocolMock('somewhere');
+
+  it('should throw an error if one tries to set unvalid properties', () => {
```
scottinet:
```suggestion
  it('should throw an error if one tries to set invalid properties', () => {
```

src/proxify.js
```diff
@@ -0,0 +1,144 @@
+/*
+Helper that can throw error when get/set undeclared properties.
```
scottinet:
```suggestion
Class wrapper throwing errors when getting/setting undeclared properties.
```

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+  seal: true,
+  sealGet: false,
+  deprecated: [],
+  exposeApi: false,
+  apiNamespace: '__proxy__',
+});
+```
+
+- `name` : the name of the proxified object for warnings
+- `seal`: does `proxify` throw error on set undeclared properties?
```
scottinet:
```suggestion
- `seal`: does `proxify` throws an error when setting undeclared properties?
```

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+- `sealGet`: does `proxify` throw error on get undeclared properties? this options is separated from `seal` because getting undefined properties can be usefull for type checking for example
```
scottinet:
```suggestion
- `sealGet`: does `proxify` throws an error when getting undeclared properties? This options is separated from `seal` because getting undefined properties can be useful for type checking for example
```

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+- `deprecated`: array of property names which produce a deprecate warning on get/set
```
scottinet:
```suggestion
- `deprecated`: array of property names producing a deprecation warning when accessed
```

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+- `warnDepreciationOnce`: only warn once per deprecated property
```
scottinet:
```suggestion
- `warnDeprecationOnce`: only warn once per deprecated property
```

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+- `exposeApi`: expose an api in the object to manipulate properties (described below)
+- `apiNamespace`: in which namespace api is exposed
+
+##### proxyfy API
```
scottinet:
```suggestion
##### proxify API
```

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+console.log(obj.bar); // throw error
+```
+*/
+
+const getOptions = options => ({
+  name: 'object',
+  seal: true,
+  sealGet: false,
+  deprecated: [],
+  warnDepreciationOnce: true,
```
scottinet:
```suggestion
  warnDeprecationOnce: true,
```

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+
+const proxify = (obj, opts = {}) => {
+  if (!obj || (typeof obj !== 'object' && typeof obj !== 'function')) {
+    throw Error('proxify only applies on non-null object');
+  }
+
+  const options = getOptions(opts);
+  const properties = ['inspect'];
+
+  const warnedDepreciation = {};
```
scottinet:
```suggestion
  const warnedDeprecation = {};
```

(to be changed everywhere obviously)

scottinet: also, you should use a `Set` instead of an object

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+          properties.push(name);
+        }
+      },
+      unregisterProps: name => {
+        const index = properties.indexOf(name);
+        if (index !== -1) {
+          properties.splice(index, 1);
+        }
+      },
+      hasProp: name => properties.includes(name)
```
scottinet: It just occured to me that this method should probably be replaced by a `has` method in the Proxy handler, allowing anyone to simply do `if (foo in bar)`

Whatever your take on whether this `hasProp` method has to go or not, I think that you should add a `has`method in the handler anyway

src/Kuzzle.js
```diff
@@ -460,20 +469,23 @@ Discarded request: ${JSON.stringify(request)}`));
       throw new Error('You must provide a valid accessor.');
     }
 
-    if (this[accessor]) {
+    if (this.__proxy__ ? this.__proxy__.hasProp(accessor) : this[accessor]) {
```
Aschen: Why checking it the object is proxified ? Since we return the proxified object in the constructor it should always be the case no?

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+      return;
+    }
+    warnedDepreciation[hash] = true;
+    // eslint-disable-next-line no-console
+    console.warn(`Warning: ${hash} is deprecated`);
+  };
+  
+  if (options.exposeApi) {
+    obj[options.apiNamespace] = {
+      registerProps: name => {
```
Aschen: Nit: for me `registerProps` mean that I can register multiple props at once like this `registerProps('foo', 'bar')`

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+        if (!properties.includes(name)) {
+          properties.push(name);
+        }
+      },
+      unregisterProps: name => {
```
Aschen: Nit: for me `unregisterProps` mean that I can register multiple props at once like this `unregisterProps('foo', 'bar')`

test/kuzzle/proxy.test.js
```diff
@@ -0,0 +1,31 @@
...
+    Http
+  } = require('../../src/protocols'),
+  ProtocolMock = require('../mocks/protocol.mock');
+
+describe('Kuzzle proxy', () => {
+  const protocolMock = new ProtocolMock('somewhere');
+
+  it('should throw an error if one tries to set invalid properties', () => {
+    const kuzzle = new Kuzzle(protocolMock);
+    should.throws(() => {
```
Aschen: (Nit for consistency reasons): usually we use the following syntax
```
should(() => {
  kuzzle.jvt = 'foobar';
}).throwError('error message')
```

src/proxify.js
thomasarbona: This syntax already works without define `has` in proxy handler

src/Kuzzle.js
thomasarbona: Because this method is used in the constructor when the object is not yet proxified

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+
+const proxify = (obj, opts = {}) => {
+  if (!obj || (typeof obj !== 'object' && typeof obj !== 'function')) {
+    throw Error('proxify only applies on non-null object');
+  }
+
+  const options = getOptions(opts);
+  const properties = ['inspect'];
+
+  const warnedDepreciation = new Set();
```
Aschen: Typo

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+
+const deleteDuplicates = arr => [...new Set(arr)];
+
+const proxify = (obj, opts = {}) => {
+  if (!obj || (typeof obj !== 'object' && typeof obj !== 'function')) {
+    throw Error('proxify only applies on non-null object');
+  }
+
+  const options = getOptions(opts);
+  const properties = ['inspect'];
```
Aschen: Could be a Set too

src/proxify.js
```diff
@@ -0,0 +1,144 @@
...
+        throw new Error(`${options.name}.${name} is not defined`);
+      }
+      if (options.deprecated.includes(name)) {
+        warnDeprecation(name);
+      }
+      return target[name];
+    },
+    set: (target, name, value) => {
+      if (options.seal && !properties.includes(name)) {
+        throw new Error(`setting a not defined '${name}' properties in '${options.name}' object`);
```
scottinet:
```suggestion
        throw new Error(`Cannot set a value to the undefined '${name}' property in '${options.name}'`);
```

PR #400, opened 2019-06-02
Fix large document search using scroll

stafyniaksacha:
## What does this PR do?
According to ES documentation, `scroll` parameter should be sent again with `_search/scroll` api call, allowing us to choose smallest timeout for scroll to live on.


Aschen:
@stafyniaksacha Can you fix the unit tests please ? I will merge it after

stafyniaksacha:
@Aschen  should be ok

PR #405, opened 2019-06-26
Add-getting-started-vuejs-without-vuex

berthieresteban:
## What does this PR do?
Add VueJS getting-started without vuex

### How should this be manually tested?
  - Step 0 : get a Kuzzle running on localhost (without anonymous restrictions)
  - Step 1 : npm install
  - Step 2 : cd doc/6/getting-started/vuejs
  - Step 3 : npm install
  - Step 4 : npm run serve-without-vuex
  - Step 5 : npm run test

### Other changes
/
### Boyscout
/

PR #414, opened 2019-07-02
add-vuejs-gs-with-vuex

berthieresteban:
## What does this PR do?
Add vuejs getting-started using vuex
Template and CSS are the same as #405 just move logical part into vuex store

  - Step 0 : get a Kuzzle running on localhost (without anonymous restrictions)
  - Step 1 : npm install
  - Step 2 : cd doc/6/getting-started/vuejs
  - Step 3 : npm install
  - Step 4 : npm run serve-with-vuex
  - Step 5 : npm run test

### Other changes
/
### Boyscout
/


doc/6/getting-started/vuejs/with-vuex/src/store.js
```diff
@@ -0,0 +1,68 @@
...
+    SUBSCRIBE_MESSAGES: async ({ commit }, { kuzzle }) => {
+      const roomID = await kuzzle.realtime.subscribe("chat", "messages", {}, notification => {
+        if (
+          notification.type === "document" &&
+          notification.action === "create"
+        ) {
+        commit('ADD_MESSAGE', notification.result);
+        }
+      });
+      commit('SET_ROOMID', roomID);
```
jenow: You set the roomId in the store but you never use it. I think you should unsubscribe this roomId at some point (when the route change for example)

berthieresteban: Effectively, I did this in the other vuejs [gettings-started](https://github.com/kuzzleio/sdk-javascript/blob/b8b2c35055a6fde2593e394572b47178acc2f7c0/doc/6/getting-started/vuejs/without-vuex/src/App.vue), and i explained why I set the roomId in comments because in this little app I don't think its better to weighing the router. 
Should I change that in both vuejs getting-started ?

jenow: I'm just worried people would not understand why we do this in the getting started so I think it would be better, yes.

berthieresteban: Okay I'll do the changes :)

berthieresteban:
@benoitvidis 
This PR is only for bring both vuejs getting started on the same branch wich is _save-vuejs-getting-started_. Of course, we mustn't merge this branch since we resolve the problem !
The index.md will follow in another PR, I don't want you to review thousands additions in one time :p

Issue #418, opened 2019-07-12
SDK 6 Http Protocol cause JS engine crash with connect on offline


ScreamZ:
I'm building an offline app as PWA.

Doing this while switching network offline or stopping kuzzle server
```ts
const kuzzle = new Kuzzle(new Http(process.env.API_HOST));

try {
   await kuzzle.connect();
 } catch (e) {
   console.log(e);
 }
```

results in `:7512/:1 GET http://localhost:7512/ net::ERR_INTERNET_DISCONNECTED`

will crash the client which is unable to catch the following error thrown by connect.

## Related

- Others methods are immunised, only `connect` seems causing issues.
- Websocket seems immunised.

Aschen:
Hi @ScreamZ ,

Thanks you for your feedback! This will be fixed by this PR: https://github.com/kuzzleio/sdk-javascript/pull/420 (explanation are in the PR description)

ScreamZ:
Closed, resolved by https://github.com/kuzzleio/sdk-javascript/pull/420

Kafigueiredo:
@Aschen when i'm offline, i want to catch an error. This code never resolve: i can't catch this error. 
```js
  try {
    await kuzzle.document.create("xxx", "xxx", xxx);
  } catch (error) {
    console.log(error);
  }
  ```
Chrome console return me : 
```
OPTIONS http://localhost:7512/xxx/xxx/_create net::ERR_INTERNET_DISCONNECTED
```

scottinet:
Hi,

There was a promise-leak occuring on network disconnection, this should be resolved by https://github.com/kuzzleio/sdk-javascript/pull/427

(not yet released on npm)

scottinet:
Hi @Kafigueiredo ,

There was a promise-leak occuring on network disconnection, this should be resolved by https://github.com/kuzzleio/sdk-javascript/pull/427

(not yet released on npm)

PR #420, opened 2019-07-12
Fix http protocol unresolved promise on connection error

Aschen:
## What does this PR do?

When the `XmlHttpRequest` fail, the associated promise is never resolved nor rejected.  
It appear that the `XmlHttpRequest` fail in another thread so the `try ... catch` block fail to catch the associated error.  

This PR add a listener on `onreadystatechange` to reject the promise when the `XmlHttpRequest` is `done` and the status is `fail` .

Fix https://github.com/kuzzleio/sdk-javascript/issues/418

  - Step 1 : Build the SDK `npm run build`
  - Step 2 : Create the following webpage at the sdk repo root:
```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>Kuzzle SDK Embedded</title>
    <script type="text/javascript" src="dist/kuzzle.js"></script>
  </head>
  <body>
    <button id="start" onclick="start()">Start</button>
    <script type="text/javascript">
      async function start() {
        window.kuzzle = new KuzzleSDK.Kuzzle(new KuzzleSDK.Http('localhost', { port: 7512}))
        window.kuzzle.on('networkError', () => console.log('Catched by listener'))

        try {
          await window.kuzzle.connect()
        } catch (error) {
          console.log("Catched by try ... catch")
          console.log(error)
        }
      }

    </script>
  </body>
</html>
```
  - Step 3 :  Open the file in your browser and then click "start"
  - Step 4 : You should see the following messages
image

### DIFF
--- a/src/protocols/http.js
+++ b/src/protocols/http.js
@@ -173,7 +173,7 @@ class HttpWrapper extends KuzzleAbstractProtocol {
 
       xhr.onreadystatechange = () => {
         if (xhr.readyState === 4 && xhr.status === 0) {
-          reject(new Error('Cannot send the request. Is the host online?'));
+          reject(new Error('Cannot connect to host. Is the host online?'));
         }
       };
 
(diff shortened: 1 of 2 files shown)


Aschen:
> Since this PR targets `master` directly, it is a hotfix. This means that you have to explain why this is a hotfix, and bump the version numbers.

@scottinet It was just a bugfix, my bad

PR #421, opened 2019-07-15
Get api routes from server:publicApi

Aschen:
## What does this PR do?

Get API URLs list from the new `server:publicApi` route instead of the static routes file.  

If the protocol can not fetch the URLs from `server:publicApi`, there is some fallback:
 - if the route is not available for anonymous user: print warning message and use static routes
 - if the route does not exists (Kuzzle version < 1.9.0): get routes from `server:info`
    - if `server:info` is restricted: print warning message and use static routes

Depends on https://github.com/kuzzleio/kuzzle/pull/1350

## Other changes

  - update the static routes file


.travis.yml
```diff
@@ -82,7 +83,8 @@ jobs:
 
       before_script:
         - npm run doc-prepare
-        - npm run --prefix doc/framework clone-repos
+        - if [[ $TRAVIS_BRANCH =~ "^master|[0-9]+-stable$" ]]; then export BRANCH=stable; else export BRANCH=dev; fi
```
alexandrebouthinon: NP: readability improvement:
```suggestion
        - |
        if [[ $TRAVIS_BRANCH =~ "^master|[0-9]+-stable$" ]] then 
          export BRANCH=stable
        else 
          export BRANCH=dev
        fi
```

PR #423, opened 2019-07-16
Emit queryError event on malformed request

### LINKED ISSUE #357: [6-dev] discarded payload is treated as a discarded request
## Expected Behavior
We can't distinguish between discarded requests and responses.
There is actually only on event `discarded` which is emitted when : 
- a _query_ is emitted when [a protocol is not connected](https://github.com/kuzzleio/sdk-javascript/blob/6-dev/src/protocols/abstract/common.js#L78)
- a non queuable _query_ is discarded [during queue registration](https://github.com/kuzzleio/sdk-javascript/blob/6-dev/src/Kuzzle.js#L405)
- **but also when a [_response_ is discarded](https://github.com/kuzzleio/sdk-javascript/blob/6-dev/src/protocols/websocket.js#L92)**

It would be nice to be able to manage both cases independently.

Aschen:
## What does this PR do?

Emit a `queryError` when receiving a response from a malformed request

Fix #357

### DIFF
--- a/doc/6/essentials/events/index.md
+++ b/doc/6/essentials/events/index.md
@@ -86,7 +86,7 @@ Triggered whenever Kuzzle responds with an error
 **Callback arguments:**
 
 `@param {KuzzleError} error - Error details`
-`@param {object} request - Request that causing an error`
+`@param {object} request - Request that caused the error`
 
 ## reconnected
 
--- a/doc/6/essentials/events/index.md
+++ b/doc/6/essentials/events/index.md
@@ -85,12 +85,8 @@ Triggered whenever Kuzzle responds with an error
 
 **Callback arguments:**
 
-`@param {object} data`
-
-| Property  | Type              | Description                   |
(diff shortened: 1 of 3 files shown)


stafyniaksacha:
Hum, I think the discarded event should not be raised here, as it is not the same event type

doc/6/essentials/events/index.md
```diff
@@ -85,12 +85,8 @@ Triggered whenever Kuzzle responds with an error
...
 **Callback arguments:**
 
-`@param {object} data`
-
-| Property  | Type              | Description                   |
-| --------- | ----------------- | ----------------------------- |
-| `request` | <pre>object</pre> | Request that causing an error |
-| `error`   | <pre>Error</pre>  | Error details                 |
+`@param {KuzzleError} error - Error details`
+`@param {object} request - Request that causing an error`
```
scottinet:
```suggestion
`@param {object} request - Request that caused the error`
```

PR #424, opened 2019-07-16
Prevent pending request leak when disconnect the SDK

### LINKED ISSUE #356: [6-dev] event listeners leak on protocol disconnect
## Current Behavior
When a query is emitted by a protocol, an event listener is registered to handle the server answer (see [here in abstract protocol](https://github.com/kuzzleio/sdk-javascript/blob/6-dev/src/protocols/abstract/common.js))

The problem is that nothing discard pending, in memory, listeners when something wrong append with network

## Possible Solution
Once listeners should be cleaned when a disconnection occurred. 
It would be nice to let user to choose what to do at this point (replay request / silent discard / error request)

Aschen:
## What does this PR do?

When a request is made to Kuzzle, we setup a once listener to handle the response. If a disconnection occur (manual with `disconnect()` or not), the listener are never freed.  

This PR save the pending requests in a `Map` and when a disconnection occur, a `discarded` event is triggered for each requests and then the map is cleared.

Fix #356 

### Other changes

  - add .npmignore

### DIFF
--- a/src/protocols/abstract/common.js
+++ b/src/protocols/abstract/common.js
@@ -23,8 +23,6 @@ class AbstractWrapper extends KuzzleEventEmitter {
         this[opt] = options[opt];
       }
     });
-
-    this.on('disconnected', () => this.clear());
   }
 
   get host () {
@@ -68,6 +66,8 @@ class AbstractWrapper extends KuzzleEventEmitter {
    * Called when the client's connection is established
    */
   clientConnected (state, wasConnected) {
+    this.on('disconnected', () => this.clear());
+
     this.state = state || 'ready';
     this.emit(wasConnected && 'reconnect' || 'connect');
   }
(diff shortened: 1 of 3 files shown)


src/protocols/abstract/common.js
```diff
@@ -110,6 +116,18 @@ Discarded request: ${JSON.stringify(request)}`));
     return this.state === 'ready';
   }
 
+  /**
+   * Clear pendings requests.
+   * Emits an event for each discarded pending request.
+   */
+  clear () {
+    for (const request of this._pendingRequests.values()) {
```
Yoann-Abbes: why not forEach?

Aschen: I try to avoid `forEach` because it's error prone when dealing with promise. I could use it here but I prefer to stick to the same construct everywhere

test/protocol/common.test.js
```diff
@@ -50,11 +40,21 @@ describe('Protocol query management', () => {
       const request = {requestId: 'bar', response: {}};
 
       protocol.query(request);
+
       should(sendSpy)
         .be.calledOnce()
         .be.calledWith(request);
     });
 
+    it('should adds the requests to pending requests', () => {
```
scottinet:
```suggestion
    it('should add the requests to pending requests', () => {
```

PR #427, opened 2019-07-19
Solve promise+event+memory leaks when the network fails

scottinet:
# Description

Whenever a request is sent through `AbstractProtocol.query`:
* a one-time listener is registered to the requestId event
* a new entry is created in a map holding pending requests
* a promise is created and returned to the caller

When an API response is received, its requestId is used to invoke the one-time listener which, in turns, remove the entry from the pending requests map, and resolves/rejects the promise.

But nothing was ever done if the network fails, meaning that requests were created with no hope of ever receiving a corresponding response.
This leads to several leaks, and it will also block users awaiting for SDK methods that might never resolve their returned promise.

This PR fixes the problem by making sure that `AbstractProtocol.clear` gets called on network disconnections (expected or not), ensuring that the pending requests map is properly cleared.
This PR also enhances the "clear" method by making it unregister the one-time event, and reject the corresponding promise, to make sure that no trace remains of the pending request.

### DIFF
--- a/test/protocol/common.test.js
+++ b/test/protocol/common.test.js
@@ -1,8 +1,11 @@
+const root = '../..';
+
 const
   should = require('should'),
   sinon = require('sinon'),
-  KuzzleError = require('../../src/KuzzleError'),
-  AbstractWrapper = require('../../src/protocols/abstract/common');
+  KuzzleError = require(`${root}/src/KuzzleError`),
+  AbstractWrapper = require(`${root}/src/protocols/abstract/common`),
+  PendingRequest = require(`${root}/src/protocols/abstract/pendingRequest`);
 
 describe('Common Protocol', () => {
   let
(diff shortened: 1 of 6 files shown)


test/protocol/common.test.js
```diff
@@ -1,8 +1,11 @@
+const root = '../..';
+
 const
   should = require('should'),
   sinon = require('sinon'),
-  KuzzleError = require('../../src/KuzzleError'),
-  AbstractWrapper = require('../../src/protocols/abstract/common');
+  KuzzleError = require(`${root}/src/KuzzleError`),
```
Aschen: This kind of shortcut beak editor ability to automatically refactor imports when renaming files

scottinet: I wouldn't know, I don't use IDEs :expressionless: 

I'll revert that one, since this makes code marginally more readable.

PR #437, opened 2019-08-01
Fix duplicate subscriptions on reconnect

benoitvidis:
# What does this PR do?

When the jwt expires, all subscriptions are reset. Users need to manually resubscribe on the `tokenExpired` event.

This PR attempts to fix the following situations:

- If we resubscribe on `tokenExpired` and `autoResubscribe` is on and if Kuzzle client gets disconnected and the token expires during the disconnection, when the client is reconnected, the subscriptions are added twice.
- Some `tokenExpired` collisions can mess the attached events
- `UnhandledRejection` issued by a failing autoresubscribe if the token expired in the meantime

# Changes

- The `realtime` controller cleans up the rooms on `disconnect` and `networkError` events to avoid some leaks.
- The `tokenExpired` event is now handled at Kuzzle level and
  1. removes all subscriptions even if coming from the protocol layer
  2. throttles to avoid duplicates
- Failing auto-re-subscriptions are now caught and trigger a `discarded` event.


benoitvidis:
Keeping in wip for the time being:
- some tests are missing
- [This snippet](https://gist.github.com/benoitvidis/04aebc0a316f7487f638ee3ca5495248) looses its subscriptions randomly after the token expires.

benoitvidis:
Removing wip: The attached gist failure is due to a bug in Kuzzle and addressed [here](https://github.com/kuzzleio/kuzzle/pull/1382)

benoitvidis:
@scottinet You're right. The fixed behaviours have been there for a while now. I think we can wait for the next release.

PR #438, opened 2019-08-02
Fix functional tests and run eslint on them

scottinet:
# Description

An error message had a slight modification in Kuzzle, and this breaks a functional test on this project because we made an exact match.

This PR makes functional tests work again.

# Other changes

* eslint now scans the `featues` directory
* fixes eslint errors and warnings on the `features` directory


features/steps/document.js
```diff
@@ -265,10 +263,8 @@ When('I update the documents [{string}, {string}]', async function (id1, id2) {
 
 
 Then('I get an error with message {string}', function (message) {
-  should(this.error)
-    .not.be.null();
-  should(this.error.message)
-    .eql(message);
+  should(this.error).not.be.null();
+  should(this.error.message).startWith(message);
```
scottinet: This is the actual functional tests fix.

benoitvidis: As the error message change will impact all sdks, I would rather have updated the submodule reference (the error message is already modfiied on the HEAD of `sdk-features`)

scottinet:
@benoitvidis > of course you're right... I don't even now what went through my head on that one :man_facepalming: 

I will do the following:
1. revert the change made on the error message check
2. rebase this PR toward `6-dev` and simply make it a PR about eslint
3. write a new PR fixing the SDK features repository

Sorry, this is an embarassing mistake. :disappointed:

scottinet:
... and you already updated the SDK features repository yesterday. :expressionless:

PR #441, opened 2019-08-09
Fix base64 decoding for the JWT in React Native

Aschen:
## What does this PR do?

React Native does not have the `Buffer` class or the `atob` function to decode base64 payload.  
This PR add a standalone base64 decoding function to ensure portability across platforms.

  - Step 1 : Run unit tests

### DIFF
--- a/src/core/Jwt.js
+++ b/src/core/Jwt.js
@@ -10,7 +10,7 @@ const decodeBase64 = input => {
   let output = '';
 
   if (str.length % 4 === 1) {
-    throw new Error('Malformated base64 string.');
+    throw new Error('Malformed base64 string.');
   }
 
   for (let bc = 0, bs = 0, buffer, i = 0;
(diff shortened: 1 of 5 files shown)


src/core/Jwt.js
```diff
@@ -1,15 +1,29 @@
...
-  if (bufferAvailable()) {
-    return Buffer.from(base64, 'base64').toString();
+const base64Charset = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=';
+
+const decodeBase64 = input => {
+  const str = input.replace(/=+$/, '');
+  let output = '';
+
+  if (str.length % 4 === 1) {
+    throw new Error('Malformated base64 string.');
```
scottinet:
```suggestion
    throw new Error('Malformed base64 string.');
```

PR #450, opened 2019-10-07
Remove local checks for arguments

Aschen:
## What does this PR do?

Remove the local check of the arguments since now we have specifics errors code sent by Kuzzle.

### Other changes

 - Add `id` and `code` properties to `KuzzleError`
 - Allow custom `sdkName` and `sdkInstanceId` in the volatile


.ci/test-docs.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+set -e
+
+here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd "$here"
+
+# docker-compose -f ./doc/docker-compose.yml pull
```
scottinet: Either uncomment or delete these lines.

PR #451, opened 2019-10-07
Remove local checks for arguments 

Aschen:
## What does this PR do?

`v6` version of #450 

Remove the local check of the arguments since now we have specifics errors code sent by Kuzzle.

### Other changes

 - Add error codes `id` and `code` properties to `KuzzleError`


Aschen:
> Did you test passing null values or empty strings to functions?

No but Kuzzle will respond with an error if one of the parameter is incorrect

PR #453, opened 2019-10-10
Fix connected property

Aschen:
## What does this PR do?

The `kuzzle.connected` property which delegate to the protocol was testing a wrong `protocol.state` value (`'online'` instead of `'connected'`). 

6.x PR: https://github.com/kuzzleio/sdk-javascript/pull/452


Aschen:
I will fix functional tests in a separate PR

PR #436, opened 2019-08-01
Add getting started React

berthieresteban:
## What does this PR do?
Add Kuzzle React getting-started.

This getting-started have been merge in the last doc [#291](https://github.com/kuzzleio/documentation/pull/291)

The last getting started app was builted progressively and some code was rewrite during the reading. 
I had to rewrite lot of parts to remove that because it implicated some untested lines. I avoid hard snippets written in the readme too.

I removed the tested snippets and add cypress to test the app globally.

The app use the current 6-dev sdk version with the ugly post-install line in `doc/6/getting-started/.react/package.json`

### How should this be manually tested?

  - Step 0 : Kuzzle running stack
  - Step 1 : cd doc/6/getting-started/.react
  - Step 2 : yarn install
  - Step 3 : yarn start &  (if error add `SKIP_PREFLIGHT_CHECK=true`)
  - Step 4 : npm run test

### Other changes
/

### Boyscout
/

PR #431, opened 2019-07-29
Add getting started Vue.js

berthieresteban:
## What does this PR do?

This getting-started has been added to the documentation before but removed because vuepress can't build with it. 

It was approved in those PR : #405 and #406 

Add VueJS getting started without vuex to SDK JS6 documentation.
Resolve the build problem of the vuepress doc (Vuepress tried to build .vue files of the getting-started and crash) by adding symlink.

image

:warning:  need to add following line in doc/6/getting-started/vuejs/standalone/index.md line 14 after merge: 
``` You can find the full code of this guide [here](https://github.com/kuzzleio/sdk-javascript/tree/6-dev/doc/6/getting-started/vuejs/standalone).```:warning: 

### How should this be manually tested?
See documentation: 
  - Step 1 : npm run doc prepare
  - Step 2 : npm run doc-dev

Test getting started project (You'll need the CYPRESS_RECORD_KEY as environment variable)
  - Step 1 : Get a Kuzzle running without anonymous restrictions
  - Step 2 : cd doc/6/getting-started/vuejs && npm ci
  - Step 3 : npm run serve-standalone
  - Step 4 : npm run test

### Other changes
/

### Boyscout
/

PR #454, opened 2019-10-11
Import Vue.js and React getting started from v6

scottinet:
# Description

Import #431 and #436 to v7: commits were cherry-picked, then the getting started directories were moved to the `doc/7` documentation directory. Finally, the `test-doc` script has been adapted to run tests on the v7 version of the getting started.

Those are the only changes made, there should be no need to review the full content of this PR (but you may do it if you want to)


Aschen:
Test will be fixed in a different PR

PR #456, opened 2019-10-25
Do not try to reconnect if the browser is offline

scottinet:
# Description

If `autoReconnect` is true, the SDK tries to reconnect every `reconnectionDelay` milliseconds.

If executed within a browser, the SDK can know if the browser has access to some kind of network: if it's marked as "offline", then there is no point retrying to connect, and this might even be harmful in some situations. For instance, browsers switch to offline if a laptop lid is closed, or if a mobile phone screen is turned off. So continuing to connect is just a useless consumption of battery power.

This PR stops the reconnection loop and instead waits for the browser to switch to "online" again before resuming its reconnection attempts.

### DIFF
--- a/src/protocols/abstract/realtime.js
+++ b/src/protocols/abstract/realtime.js
@@ -71,7 +71,7 @@ class RTWrapper extends KuzzleAbstractProtocol {
           'online',
           () => {
             this.retrying = false;
-            this.connect(this.host).catch(err => this.clientNetworkError(err));
+            this.connect().catch(err => this.clientNetworkError(err));
           },
           { once: true });
         return;
--- a/src/protocols/abstract/realtime.js
+++ b/src/protocols/abstract/realtime.js
@@ -4,7 +4,6 @@ const
   KuzzleAbstractProtocol = require('./common');
 
 class RTWrapper extends KuzzleAbstractProtocol {
-
   constructor (host, options = {}) {
     super(host, options);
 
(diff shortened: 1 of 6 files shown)


src/protocols/abstract/realtime.js
```diff
@@ -52,27 +51,43 @@ class RTWrapper extends KuzzleAbstractProtocol {
...
 
+      if ( typeof window === 'object'
+        && typeof window.navigator === 'object'
+        && window.navigator.onLine === false
+      ) {
+        window.addEventListener(
+          'online',
+          () => {
+            this.retrying = false;
+            this.connect(this.host).catch(err => this.clientNetworkError(err));
```
Aschen:
```suggestion
            this.connect().catch(err => this.clientNetworkError(err));
```

scottinet:
@Aschen > About this PR's goal: assuming the SDK is run in a browser, if `onLine` is false, then we know for certain that the navigator don't have access to any kind of network, so we stop the reconnection loop until the navigator goes online again, potentially saving battery life on laptops, mobiles, ...
But the reverse is not true: a navigator `onLine` status set to true does not necessarily means that network is available (e.g. the device might be connected to a restricted, or badly configured, or faulty network). 
So we don't rely on it entirely: 
* we start the reconnection loop once the TCP socket is closed (which should happen _before_ the "online" event is triggered anyway)
* once the navigator goes "online" again, we don't reconnect directly. Instead, we restart the reconnection loop, because we don't know for sure that we'll be able to reach the kuzzle server

So we never need to listen to the "online" event.

PR #464, opened 2019-11-14
Fix option refresh in queries

Yoann-Abbes:
## What does this PR do?

This PR fixes the way we pass the `refresh` option in several routes.
Before that, the option was lost and never passed to ES.

Ticket [Jira](https://jira.kaliop.net/projects/KZL/issues/KZL-1402?filter=allopenissues)

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...

### DIFF
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -407,7 +407,7 @@ class Kuzzle extends KuzzleEventEmitter {
     }
 
     // we follow the api but allow some more logical "mistakes" (the only allowed value for refresh arg is "wait_for")
-    if (request.refresh) {
+    if (request.refresh || options.refresh) {
       request.refresh = 'wait_for';
     }
 
--- a/src/Kuzzle.js
+++ b/src/Kuzzle.js
@@ -407,8 +407,9 @@ class Kuzzle extends KuzzleEventEmitter {
     }
 
     // we follow the api but allow some more logical "mistakes" (the only allowed value for refresh arg is "wait_for")
-    if (request.refresh || options.refresh) {
+    if (options.refresh) {
       request.refresh = 'wait_for';
+      delete options.refresh;
     }
 
     if (!request.volatile) {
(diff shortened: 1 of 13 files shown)


src/controllers/collection.js
```diff
@@ -12,21 +12,28 @@ class CollectionController extends BaseController {
...
+    const request = {
       index,
       collection,
       body: mappings,
-      action: 'create'
-    }, options)
+      action: 'create',
+      refresh: options.refresh
+    };
+    delete options.refresh;
```
scottinet: ..why, since now Kuzzle.query handles the refresh option?

It seems to me that you don't need to change controllers to fix the issue.

Yoann-Abbes: I changed the way it fix the issue in the last commit

test/controllers/collection.test.js
```diff
@@ -28,7 +28,8 @@ describe('Collection Controller', () => {
               action: 'create',
               body: null,
               index: 'index',
-              collection: 'collection'
+              collection: 'collection',
+              refresh: undefined
```
scottinet: Is this necessary?

src/Kuzzle.js
```diff
@@ -407,7 +407,10 @@ class Kuzzle extends KuzzleEventEmitter {
     }
 
     // we follow the api but allow some more logical "mistakes" (the only allowed value for refresh arg is "wait_for")
-    if (request.refresh) {
+    if (request.refresh || options.refresh) {
+      if (options.refresh) {
+        delete options.refresh;
```
Aschen: Why do you delete the property after using it? 
It can be confusing to mutate the original object, it could be reused for subsequent calls:
```
const options = { refresh: 'wait_for' };
sdk.query({...}, options);
sdk.query({...}, options); // options object have been mutated 
```

PR #465, opened 2019-11-15
Choose GET route by default in case of double route

Yoann-Abbes:
## What does this PR do?

This PR changes the way we get the correct route if there is several of them.

Before this PR:
The SDK checked the size of the url, and chose the shortest.
If they had the same size, it chose the POST route.
And you could not force the verb in your request.

Now:
Same thing, except it chooses the GET route.
Plus, you can force the verb with the option `{verb: 'POST'}`

Install this version of the sdk `npm install git://github.com/kuzzleio/sdk-javascript#handle-double-routes`
Run Kuzzle in debug mod.
Run this script
```
const { Kuzzle, Http } = require('kuzzle-sdk');

const kuzzle = new Kuzzle(
  new Http(`localhost`, { port: 7512})
);

async function main () {
const doc1 = { capacity: 4 };
const doc2 = { capacity: 7 };

    try {
	await kuzzle.connect();
	await kuzzle.document.createOrReplace('nyc-open-data', 'yellow-taxi', 'some-id', doc1);
	await kuzzle.document.createOrReplace('nyc-open-data', 'yellow-taxi', 'some-other-id', doc2);

  const response = await kuzzle.document.mGet(
    'nyc-open-data',
    'yellow-taxi',
      ['some-id', 'some-other-id']
//   ,{verb: 'POST'}
  );

  console.log(response);

} catch (error) {
  console.error(error.message);
}
}

main()

```

You will see in the logs that the route verb was GET

Now uncomment the line `//   ,{verb: 'POST'}`
The route will be in POST.

You can do the same with `mGetUsers`

### Other changes

Adapt doc snippet test for `getAllStats` and `getLastStats`, because of the new properties order of the response object [recently defined](https://github.com/kuzzleio/kuzzle/pull/1561/files#diff-ed4e16c2c6915728845ce78457bf7c9dL44)


Yoann-Abbes:
I've tested it by running Kuzzle with this branch https://github.com/kuzzleio/kuzzle/pull/1519
It works and tests are ok

src/controllers/document.js
```diff
@@ -136,10 +136,15 @@ class DocumentController extends BaseController {
     const request = {
       index,
       collection,
-      body: {ids},
-      action: 'mGet'
+      action: 'mGet',
+      includeTrash: options.includeTrash
```
Aschen: This options does not exists in Kuzzle v2 anymore, it was related to the trashcan which have been removed

src/controllers/document.js
```diff
@@ -136,10 +136,15 @@ class DocumentController extends BaseController {
...
     };
 
+    if (options.verb === 'POST') {
+      request.body = {ids};
+    } else {
+      request.ids = ids.toString();
```
Aschen: (Nitpicking) Prefer `Array.join` because it's more idiomatic. Also, the `else` is not indented regarding our coding standards

doc/7/controllers/security/m-get-users/index.md
```diff
@@ -0,0 +1,41 @@
...
+---
+
+# mGetUsers
+
+Gets multiple security users.
+
+<br />
+
+```js
+mGetUsers(ids, [options]);
```
Aschen:
```suggestion
mGetUsers(kuids, [options]);
```

doc/7/controllers/security/m-get-users/index.md
```diff
@@ -0,0 +1,41 @@
...
+### options
+
+| Property | Type<br />(default) | Description |
+| --- | --- | --- |
+| `queuable` | <pre>boolean</pre><br />(`true`) | If true, queues the request during downtime, until connected to Kuzzle again |
+| `verb`     | <pre>string</pre>      | Forces the verb of the route |
+
+## Resolves
+
+An array of retrieved [`User`](/sdk/js/6/core-classes/user/introduction) objects.
```
Aschen:
```suggestion
An array of retrieved [`User`](/sdk/js/7/core-classes/user/introduction) objects.
```

doc/7/controllers/security/m-get-users/index.md
```diff
@@ -0,0 +1,41 @@
...
+Gets multiple security users.
+
+<br />
+
+```js
+mGetUsers(kuids, [options]);
+```
+The route exists in `GET` and `POST`.
+By default, the SDK hits the `GET` one.
+You can force it to be `POST` in the `options`.
```
scottinet: This shouldn't be part of the route description, but should instead be a dedicated sub-chapter for the `verb` option below.

Something such as:

```
#### verb

When instantiated with a HTTP protocol object, the SDK uses the GET API by default for this API route.
You can set the `verb` option to `POST` to force the SDK to use the POST API instead.
```

doc/7/controllers/security/m-get-users/index.md
```diff
@@ -0,0 +1,41 @@
...
+|--- |--- |--- |
+| `ids` | <pre>array&lt;string&gt;</pre> | User identifiers |
+| `options` | <pre>object</pre> | Query options |
+
+### options
+
+| Property | Type<br />(default) | Description |
+| --- | --- | --- |
+| `queuable` | <pre>boolean</pre><br />(`true`) | If true, queues the request during downtime, until connected to Kuzzle again |
+| `verb`     | <pre>string</pre>      | Forces the verb of the route |
```
scottinet:
```suggestion
| `verb`     | <pre>string</pre>      | (HTTP only) Forces the verb of the API route |
```

doc/7/controllers/security/m-get-users/index.md
```diff
@@ -0,0 +1,41 @@
...
+```
+The route exists in `GET` and `POST`.
+By default, the SDK hits the `GET` one.
+You can force it to be `POST` in the `options`.
+
+<br />
+
+| Property | Type | Description |
+|--- |--- |--- |
+| `ids` | <pre>array&lt;string&gt;</pre> | User identifiers |
```
scottinet: our standard is more like this:

```suggestion
| `ids` | <pre>string[]</pre> | User identifiers |
```

(see [this route](https://docs.kuzzle.io/sdk/js/7/controllers/document/m-create/) for instance)

src/controllers/document.js
```diff
@@ -194,6 +194,7 @@ class DocumentController extends BaseController {
       action: 'search',
     };
 
+    options.verb = 'POST';
     for (const opt of ['from', 'size', 'scroll']) {
       request[opt] = options[opt];
       delete options[opt];
```
Aschen: I know it's not directly related to this PR but here the `options` object should not be mutated because the user may want to re-use it across subsequent calls

Yoann-Abbes:
@Aschen despite our discussions, everything is OK

PR #479, opened 2020-01-09
Add a "source" option for document:update to retrieve the updated document

Yoann-Abbes:
## What does this PR do?

This PR adds a `source` option. Set to `true` returns the updated document in a `_source` object.

Depends to https://github.com/kuzzleio/kuzzle/pull/1549

Run Kuzzle within this branch https://github.com/kuzzleio/kuzzle/pull/1549

`document.update(index, collection, body, {source: true})`


Yoann-Abbes:
>  LGTM but missing a new end to end test to cover this use case

Functional test are in Kuzzle (which use the SDK)

Aschen:
IMHO, unit tests are enough for a small improvement like this

PR #485, opened 2020-01-24
Implement global timeout

Aschen:

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...


src/protocols/http/abortableRequest.js
```diff
@@ -0,0 +1,166 @@
...
+];
+
+class AbortableRequest {
+  constructor (protocol, timeout, options) {
+    this._protocol = protocol;
+    this._timeout = timeout;
+    this._options = options;
+    this._request = null;
+
+    this._options.timeout = this._timeout;
```
rolljee: may options be undefined or null ?
shall we `constructor (protocol, timeout, options = {}) {}`
To avoid potential issue while dealing with nested data inside this object ? 
error

PR #489, opened 2020-02-08
Allow to modify SDK actions behavior with pipes

Aschen:
## What does this PR do?

This PR adds the ability to modify SDK actions behavior by registering pipe functions.  

The Kuzzle object now expose a [registerPipe](https://deploy-preview-489--doc-sdk-javascript.netlify.com/core-classes/kuzzle/register-pipe/) method. 

This method allow to execute pipes to modify the action payload (`before`) or the action result (`after`). Each method should return the same payload either directly or through a promise resolving to it.

For now, only the `query` method can be modified with pipes:
 - `kuzzle:query:before`: modify the `request` before being sent to Kuzzle
 - `kuzzle:query:after`: modify the response sent by Kuzzle

Example:
```js
kuzzle.registerPipe('kuzzle:query:after', 'parse server:now', response => {
  if (response.controller !== 'server' || response.action === 'now') {
    return;
  }

  response.result.parsed = new Date(response.result.now);
});
```

Related documentation:
  - [Extend SDK page](https://deploy-preview-489--doc-sdk-javascript.netlify.com/essentials/extend-sdk/register-pipe/)
  - [Kuzzle.registerPipe](https://deploy-preview-489--doc-sdk-javascript.netlify.com/core-classes/kuzzle/register-pipe/)

### Other changes

 - authentication token addition to the request is now done with a pipe


scottinet:
What's the use case?

The way I see it, SDK actions are all pretty much atomic, so there isn't really room for interleaving.

To take your example, what is the advantage of pipes over something like that?

```js
async function typedNow() {
  const response = await kuzzle.server.now();
  return new Date(response.now);
}
```

Aschen:
> What's the use case?
> 
> The way I see it, SDK actions are all pretty much atomic, so there isn't really room for interleaving.
> 
> To take your example, what is the advantage of pipes over something like that?
> 
> ```js
> async function typedNow() {
>   const response = await kuzzle.server.now();
>   return new Date(response.now);
> }
> ```

The first usecase for registering a pipe before sending a query is to add the JWT to the request: https://github.com/kuzzleio/sdk-javascript/pull/489/files#diff-0e9dabac598da6abebd4ae2d2a2c3c24R23

You may want to modify on-the-fly request before being sent to Kuzzle, for example with the multi-tenant plugin it could be useful to modify user request to the document controller by injecting the index he is allowed to read/write.
A pipe on the client side is more precise than a pipe on Kuzzle side because you can setup pipes separately for each SDK instantiation.

scottinet:
Ah, clever use of that feature for JWT injection.

But I'm a bit uneasy regarding this:

> You may want to modify on-the-fly request before being sent to Kuzzle, for example with the multi-tenant plugin it could be useful to modify user request to the document controller by injecting the index he is allowed to read/write.

Because the SDK API makes it clear that it requires an `index` and a `collection`, and this example you give is cleary disturbing for users, as it makes implicit (and hidden) changes to an otherwise very explicit API.
Say you have such a pipe overwriting the index with, say,  `indexA`, and somewhere in the code you ask the SDK to create a document on  `indexB`... good luck finding what's going on if you try to debug this on a moderate to heavy front-end app : your code cleary states that you work on indexB, but if you analyze Kuzzle logs or network packets, you see that the document gets created on indexA.

For now, I still fail to see use cases where this feature would come in handy... and I'm even starting to think that it might be harmful.

Unrelated to this discussion, but since you mention plugins : since SDK operations are atomic (or nearly so), why instead of plugins don't we have layers? It would make more sense (to me) if we had a multi-tenant SDK built over the raw JS SDK, which 1. would only load parts of the SDK related to documents (to make it lighter) and 2. would expose methods designed especially for multi-tenancy, invoking the underlying SDK?
(you couldn't have multiple plugins with that kind of system however)

Aschen:
Hmm yes you're right, it could be disturbing to users to have such changes made without them knowing.  

For now I also fail to find other use case as useful as the auth token injection.. (EDIT: I found another use case by writing this response :sweat_smile: ) But IMHO it's a good thing to allow users to interact with Kuzzle request and response at a low level because in this way they can extend the SDK in ways with didn't think about.

For the multi tenant, I think the easiest way is to use a new controller (`Kuzzle.useController`) which exposes only the document controller wrapped with the correct index. But since the correct index name must be retrieved from the logged user, I don't really see how this controller could automatically adds the index name to every request. 
It would need a system to retrieve the logged user `_id` and then get the related index name, and this could be done with the `Kuzzle.registerPipe` method for example.

Aschen:
Closing this since there is not numerous concrete usecases.

Aschen:
I re-open this PR because I found another use-case: https://github.com/kuzzleio/sdk-javascript/pull/489

I need to pass an option to the `realtime.subscribe` method inside the plugin (the `cluster` option) and I could have added it with this API (and even change the default value just for the plugin SDK because plugin subscription want `cluster: false` most of time time when client subscription need `cluster: true`

scottinet:
About your need, why not implement a custom SDK in Kuzzle Core, inheriting the SDK?

Something like this:

```js
const { Kuzzle: KuzzleSDK } = require('kuzzle-sdk');

class CoreSDK extends KuzzleSDK {
  /**
  * @override
  */
  query (request = {}, options = {}) {
    // add core-specific options here
    return super.query(request, options);
  }
}
```

Aschen:
Yes it could be.
(discussion) So do you think it's better to offer adaptability through inheritance than a pipe system?

scottinet:
This depends on the scope of this feature. If this is only for our internal needs, then a wrapper seems more suited than pipes. 
If this is also useful for users, then pipes seem better. But in that case, I stated a few concerns at the beginning of this PR, and we need to address them.

Aschen:
After some thought I think the problem is more the lack of dependency/logic injection inside the controllers method and internal classes.  

For example the problem with the realtime SDK PR on the core was the Room class that doesn't follow unknown new options to the realtime.subscribe method.

Maybe instead of the pipe system we could just allow dependency injection for core classes (and also expose them in the package)

scottinet:
I like dependency injection.

Aschen:
I close this PR again and I think I'm not gonna re-open it since we agreed that dependency injection is a better way to allows user to modify the SDK behavior

PR #490, opened 2020-02-13
Throw an error when an URL param is missing in HTTP

Aschen:
## What does this PR do?

When you send a request with arguments that are present in the URL (eg: `/:index/_create`), they are injected in the URL.  
If the user didn't put the required arguments, they are `undefined` and they end up in the url as the string `"undefined"` (eg: `/undefined/create`).    

This PR now discard the request if an URL param is missing.


src/protocols/http.js
```diff
@@ -175,9 +175,10 @@ class HttpWrapper extends KuzzleAbstractProtocol {
         && this.routes[payload.controller][payload.action];
 
     if (! route) {
-      const error =
-        new Error(`No URL found for "${payload.controller}:${payload.action}".`);
-      this.emit(payload.requestId, {status: 400, error});
+      const error = new Error(
```
scottinet: (nitpicking) no need to wrap line here. One other occurrence below .

PR #492, opened 2020-02-21
Fix getting started links

berthieresteban:
## What does this PR do?
Fix links for getting started react, react native, vuejs for sdk JS 7.
Les liens vers `/core/2` ne seront pas accessible depuis le build netlify qui ne charge que les pages de doc du sdkl
Le test concernant les deadlink est pété d'ou le fail 
(ex: src/controllers/auth/login in src/sdk/js/7//getting-started/react-native/index.md)
ce lien fonctionne sur netlify

Netlify

### Other changes
/

### Boyscout
Remove link to react kuzzle boilerplate in getting started react because it is not up to date with kuzzle versions

### DIFF
--- a/doc/7/getting-started/react-native/index.md
+++ b/doc/7/getting-started/react-native/index.md
@@ -9,7 +9,7 @@ order: 400
 
 # Getting Started with Kuzzle and React Native
 
-This section deals with **Kuzzle V2** (+ **Javascript SDK 7**) and **React Native**. We will create **documents** in Kuzzle and subscribe to [document notifications](/sdk/js/7/essentials/realtime-notifications) to develop a realtime chat.
+This section deals with **Kuzzle V2** (+ **Javascript SDK 7**) and **React Native**. We will create **documents** in Kuzzle and subscribe to [document notifications](/sdk/js/7/essentials/realtime-notifications#document-messages) to develop a realtime chat.
 
 
 ## Requirements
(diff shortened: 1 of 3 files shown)


rolljee:
After CI is fixed

Aschen:
This will not fix the issue because as far I can see, we will still have problem with the dead link check.

The CI will never pass because those links are not in the same format as the others (`/sdk/js/7` or `/core/2`) and so the program can not check if they really exists.

Why Vuepress is adding `/sdk/js/7` at the beginning of links in those getting started?

berthieresteban:
This pr fix the links in the getting started. 
I agree that we still have a problem with the deadlinks check in the CI...
I don't know why Vuepress is adding `/sdk/js/7` only in the getting started links. 
Also Vuepress add this base_url only if the link start with a `/sdk/js/7` wich will become `/sdk/js/7/sdk/js/7`. 
I havn't seen any modification on the doc repository wich could explain that...

berthieresteban:
Links are fixed, and deadlinks checks are passing.
I locally verified the links and they're good. 
On Netlify, the links aren't working because the base_url wich is equal to `/` on Netlify instead of `/sdk/js/7` normally. 
For example: `/sdk/js/7/controllers/index/create` is a good link for localhost/docs.kuzzle.io but not good for preview netlify because it should be `/controllers/index/create`.

In short, links are good but leads to 404 on Netlify.

PR #498, opened 2020-03-30
Hotfix document search http

Aschen:
## What does this PR do?

When we have multiple choose for a route, we choose the shortest or the GET route.  
With `document:create` we have the choice between those 2 routes:
```
{ verb: 'GET', url: '/:index/:collection/' },
{ verb: 'POST', url: '/:index/:collection/_search' }
```

The problem is that after even if we want to change the verb by using `options.verb`, the URL is not the good one and `GET /:index/:collection` correspond to the `collection:create` method.

So this PR is just a dirty fix to always use the POST route with `document:search`.  

When we are going to fully support the GET verb with query for `document:search`, maybe we can get ride of this.

See this failing job on Kuzzle: https://travis-ci.org/github/kuzzleio/kuzzle/jobs/667658863

### DIFF
--- a/src/protocols/http.js
+++ b/src/protocols/http.js
@@ -339,7 +339,7 @@ class HttpWrapper extends KuzzleAbstractProtocol {
 }
 
 function getPostRoute (routes) {
-  return routes[0].verb === 'POST' ? routes[0] : routes[1]
+  return routes[0].verb === 'POST' ? routes[0] : routes[1];
 }
 
 function getCorrectRoute (routes) {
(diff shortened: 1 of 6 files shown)


src/protocols/http.js
```diff
@@ -307,8 +307,19 @@ class HttpWrapper extends KuzzleAbstractProtocol {
 
           if (http && http.length === 1) {
             routes[controller][action] = http[0];
-          } else if (http && http.length > 1) {
-            routes[controller][action] = getCorrectRoute(http);
+          }
+          else if (http && http.length > 1) {
+            // We need this ugly fix because the document:search route can also
```
rolljee: Is it missing one word to fully understand those 2 first comment ?

PR #508, opened 2020-04-22
Fix SearchResult.next with sort/size

Aschen:
## What does this PR do?

:warning: Depends on https://github.com/kuzzleio/kuzzle/pull/1600

Fix few bugs on the `SearchResult` pagination with `sort` and `size`.  

The SDK now support every kind of ES sort:
 - string value
 - array of string
 - array of object

Errors are now thrown when:
 - the sort is invalid
 - the sort combination does not identify one item only

### Other changes

 - the `next` method now return rejected promises instead of throwing (inconsistent behavior that can lead to unhandled rejection)
 - change `_uid` by `_id` because this field does not longer exists in ES 7.x


doc/7/core-classes/search-result/next/index.md
```diff
@@ -56,19 +56,26 @@ You can restrict the scroll session maximum duration under the `services.storage
 
 If the initial search contains `sort` and `size` parameters, the `next` method retrieves the next page of results following the sort order, the last item of the current page acting as a live cursor.
 
-To avoid too many duplicates, it is advised to provide a sort combination that will always identify one item only. The recommended way is to use the field `_uid` which is certain to contain one unique value for each document.
+This strategy use Elasticsearch [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/search-request-body.html#request-body-search-search-after) parameter.
```
Leodau, line 59:
```suggestion
This strategy uses Elasticsearch [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/search-request-body.html#request-body-search-search-after) parameter.
```

doc/7/core-classes/search-result/next/index.md
```diff
@@ -20,9 +20,9 @@ next();
 
 Resolves to a `SearchResult` object, or to `null` if no more pages are available.
 
-## Throw
+## Rejects
 
-This method throws an exception if:
+This method return a rejected promise with an error if:
```
Leodau, line 25:
```suggestion
This method returns a rejected promise with an error if:
```

scottinet:
I'm releasing Kuzzle 2.1.4, tests should pass once done.

PR #514, opened 2020-05-15
Able to execute document search as GET http method

Yoann-Abbes:
## What does this PR do?
[Jira ticket](https://jira.kaliop.net/browse/KZL-1493)

This PR provides the possibility to execute a `document:search` with the GET http verb

Create a `nyc-open-data' index` and `yellow-taxi` collection.
Create some files with 2 containing field `{name:"ok"}`
Execute this code

Remove the `verb` option and/or change it to `POST` and re-execute the code

By default, `document:search` is still POST

```js
const { Kuzzle, Http } = require('kuzzle-sdk');

const kuzzle = new Kuzzle(
  new Http(`localhost`, { port: 7512})
);

async function main () {

  try {
  	await kuzzle.connect();

        const response = await kuzzle.document.search(
      'nyc-open-data',
      'yellow-taxi',
      {
        query:{
            term: {
              name: 'ok'
            }
          }
        },{verb: 'GET'}
    );

    console.log(`Successfully get ${response.total} documents`);
  } catch (error) {
    console.error(error);
  }
}

main()
```
### Other changes

update Mocha to [7.2.0](https://github.com/mochajs/mocha/releases/tag/v7.2.0)
update codecov to 3.7.0

### Boyscout

Removes an unused url who was in conflict with `collection:update` url
Fix eslint version for the documentation CI

### DIFF
--- a/src/controllers/Document.js
+++ b/src/controllers/Document.js
@@ -188,19 +188,20 @@ class DocumentController extends BaseController {
   }
 
   _search (index, collection, body = {}, options = {}) {
+    if (options.verb && options.verb.toLowerCase() === 'get') {
+      body = {query: body};
+    }
     const request = {
       index,
       collection,
       body,
       action: 'search',
     };
-
-    for (const opt of ['from', 'size', 'scroll', 'verb']) {
+    for (const opt of ['from', 'size', 'scroll']) {
       request[opt] = options[opt];
     }
 
     const opts = { verb: options.verb || 'POST', ...options };
-
     return this.query(request, opts)
       .then(response => ({ response, request, opts }));
   }
--- a/src/controllers/Document.js
+++ b/src/controllers/Document.js
@@ -1,4 +1,5 @@
 const BaseController = require('./Base');
+const HttpProtocol = require('../protocols/Http');
 const DocumentSearchResult = require('../core/searchResult/Document');
 
 class DocumentController extends BaseController {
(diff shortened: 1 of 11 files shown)


src/controllers/Document.js
```diff
@@ -195,11 +195,11 @@ class DocumentController extends BaseController {
       action: 'search',
     };
 
-    for (const opt of ['from', 'size', 'scroll']) {
+    for (const opt of ['from', 'size', 'scroll', 'verb']) {
```
Aschen, line 198: You don't need to pass the verb as an argument since the `verb` option on query will change the http verb
```suggestion
    for (const opt of ['from', 'size', 'scroll']) {
```

src/protocols/routes.json
```diff
@@ -76,10 +76,6 @@
       "verb": "POST",
       "url": "/:index/:collection/_create"
     },
-    "search": {
```
Aschen, line 79: You should add the new route

src/controllers/Document.js
```diff
@@ -188,7 +189,10 @@ class DocumentController extends BaseController {
   }
 
   _search (index, collection, body = {}, options = {}) {
-    if (options.verb && options.verb.toLowerCase() === 'get') {
+    if ( this._kuzzle.protocol instanceof HttpProtocol
```
Aschen, line 192: It would be better to have a `name` property on the protocol objects

Yoann-Abbes: yes thanks, I was wondering

scottinet:
Seriously?
You added the getter, but you still use the class "private" property directly...

Yoann-Abbes:
> Seriously?
> You did add the getter, but you still use the class "private" property directly...

sorry, went too fast

PR #516, opened 2020-05-20
Fix query string construction

Yoann-Abbes:
## What does this PR do?

Fix a bug while constructing query strings.
The string interpolation `${key}:${value}` gave as value `[Object object]` as string for nested objects instead of stringifying the whole object. 

unit tests

### Other changes

Refresh collection before execution of `deleteByQuery` snippet code
Rename the tests scripts in package.json to make them match the kuzzle ones
Fix eslint version for the documentation CI


package.json
```diff
@@ -50,6 +50,7 @@
     "eslint": "^6.8.0",
     "eslint-friendly-formatter": "^4.0.1",
     "eslint-loader": "^4.0.0",
+    "eslint-plugin-import": "^2.20.2",
```
Aschen, line 53: What is this new dependency?

Issue #518, opened 2020-06-04
Implement document:updateByQuery


jenow:
## Expected Behavior

Be able to call `document:updateByQuery` because it exists  in the api (https://docs.kuzzle.io/core/2/api/controllers/document/update-by-query/)

## Current Behavior

Does not exist.

## Possible Solution

Implement it.

PR #519, opened 2020-06-04
Add document:updateByQuery

Yoann-Abbes:
## What does this PR do?

This PR adds `document:updateByQuery` to the sdk

Issue https://github.com/kuzzleio/sdk-javascript/issues/518

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...

### Other changes

Adds a new snippet template `print-result-array.tpl.js`


.ci/doc/templates/print-result-array.tpl.js
```diff
@@ -0,0 +1,31 @@
...
+  let result;
+  try {
+    await kuzzle.connect();
+  } catch (error) {
+    console.log(`Can not connect to Kuzzle: ${error.message}`);
+  }
+  [snippet-code] finally {
+    kuzzle.disconnect();
+  }
+  for (const elem of result.successes) {
```
Aschen, line 28: This will only works with arrays returned by some of the `m*` methods. For example it wont works with document:search or document:mDelete.

You don't need to use this template also, you can just print the content of the result in JSON with newlines

src/controllers/Document.js
```diff
@@ -220,6 +220,19 @@ class DocumentController extends BaseController {
       .then(response => response.result);
   }
 
+  updateByQuery(index, collection, searchQuery, changes, options = {}) {
+    const request = {
+      index,
+      collection,
+      body: {...searchQuery, ...changes},
```
Aschen, line 227: This should not work because every properties is at the same level instead of being inside the `searchQuery` and `changes` properties

Yoann-Abbes: ?
In searchQuery you have like 
```{ query: { match: { field: "value" } } }```
and in changes
```{ changes: { field: "value2" } }``` 
so it's ok, and works

Comment (Yoann-Abbes) [reply src/controllers/Document.js:227]: Hm you say that i should wrap it ? query and changes properties?

File: .ci/doc/templates/print-result-array.tpl.js (hunk 1 above)
Comment (Yoann-Abbes) [reply .ci/doc/templates/print-result-array.tpl.js]: I think this template could be useful, so I just renamed it `print-result-successes` as it do exactly that. So no confusion for route that wont return a `successes` array.

File: .ci/doc/templates/print-result-successes.tpl.js [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+  WebSocket
+} = require('kuzzle-sdk');
+
+// Instantiates a Kuzzle client
+const
+  kuzzle = new Kuzzle(
+    new WebSocket('kuzzle', { autoReconnect: false })
+  );
+
+// Adds a listener to detect any connection problems
```
scottinet, line 13:
```suggestion
// Adds a listener to detect connection problems
```

.ci/doc/templates/print-result-successes.tpl.js
```diff
@@ -0,0 +1,31 @@
...
+kuzzle.on('networkError', error => {
+  console.error(`Network Error: ${error.message}`);
+});
+
+(async () => {
+  let result;
+  try {
+    await kuzzle.connect();
+  } catch (error) {
+    console.log(`Can not connect to Kuzzle: ${error.message}`);
```
scottinet, line 23:
```suggestion
    console.log(`Cannot connect to Kuzzle: ${error.message}`);
```

doc/7/controllers/document/update-by-query/snippets/update-by-query.js
```diff
@@ -0,0 +1,31 @@
...
+            _id: <document-id>,
+            _source: <updated document> // if source set to true
+            status: 200
+          },
+          {
+            _id: <document id>,
+            _source: <updated document> // if source set to true
+            status: 200
+          }
+        ],
```
scottinet, line 25:
```suggestion
  {
    _id: <document-id>,
    _source: <updated document> // if source set to true
    status: 200
  },
  {
    _id: <document id>,
    _source: <updated document> // if source set to true
    status: 200
  }
],
```

PR #520, opened 2020-06-09
Hide kuzzle object from console.log

Aschen:
## What does this PR do?

Use `Reflect` to hide the kuzzle object from log

### Other changes

 - Get ride of every `delete` keyword and use `Map` for RAM caches
 - Do not throw custom error when unsubscribing from unknown room (let Kuzzle trigger it's own documented error)
 - Allows to pass `_id` to `realtime.publish`

### DIFF
--- a/src/core/Room.js
+++ b/src/core/Room.js
@@ -25,12 +25,13 @@ class Room {
       collection,
       body,
       controller: 'realtime',
-      action: 'subscribe'
+      action: 'subscribe',
+      state: this.options.state,
+      scope: this.options.scope,
+      users: this.options.users,
+      volatile: this.options.volatile,
+      cluster: this.options.cluster
     };
-    for (const opt of ['state', 'scope', 'users', 'volatile']) {
-      this.request[opt] = this.options[opt];
-      delete this.options[opt];
-    }
 
     this.autoResubscribe = typeof options.autoResubscribe === 'boolean'
       ? options.autoResubscribe
@@ -38,10 +39,6 @@ class Room {
     this.subscribeToSelf = typeof options.subscribeToSelf === 'boolean'
       ? options.subscribeToSelf
       : true;
-
-    for (const opt of ['autoResubscribe', 'subscribeToSelf']) {
-      delete this.options[opt];
-    }
 
     // force bind for further event listener calls
     this._channelListener = this._channelListener.bind(this);
--- a/src/core/Room.js
+++ b/src/core/Room.js
@@ -29,9 +29,7 @@ class Room {
       state: this.options.state,
       scope: this.options.scope,
       users: this.options.users,
-      volatile: this.options.volatile,
-      // the "cluster" option is meant to be used by the embedded plugin SDK only
-      cluster: this.options.cluster
+      volatile: this.options.volatile
     };
 
     this.autoResubscribe = typeof options.autoResubscribe === 'boolean'
--- a/src/core/Room.js
+++ b/src/core/Room.js
@@ -30,6 +30,7 @@ class Room {
       scope: this.options.scope,
       users: this.options.users,
       volatile: this.options.volatile,
+      // the "cluster" option is meant to be used by the embedded plugin SDK only
       cluster: this.options.cluster
     };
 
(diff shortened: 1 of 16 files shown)


src/core/Room.js
```diff
@@ -25,12 +25,13 @@ class Room {
       collection,
       body,
       controller: 'realtime',
-      action: 'subscribe'
+      action: 'subscribe',
+      state: this.options.state,
+      scope: this.options.scope,
+      users: this.options.users,
+      volatile: this.options.volatile,
+      cluster: this.options.cluster
```
Aschen, line 33: Actual enhancement

test/core/room.test.js
```diff
@@ -61,30 +61,30 @@ describe('Room', () => {
...
-        },
-        body = {foo: 'bar'},
-        cb = sinon.stub();
+    it('should handle scope/state/users/volatile/cluster options', () => {
+      const opts = {
+        scope: 'scope',
+        state: 'state',
+        users: 'users',
+        volatile: 'volatile',
+        cluster: false
```
Aschen, line 76: And associated test

doc/7/controllers/realtime/subscribe/index.md
```diff
@@ -40,6 +40,12 @@ Additional subscription options.
 | `users`           | <pre>string</pre><br/>(`none`)  | Subscribe to users entering or leaving the room</br>Possible values: `all`, `in`, `out`, `none`          |
 | `subscribeToSelf` | <pre>boolean</pre><br/>(`true`) | Subscribe to notifications fired by our own queries                                                      |
 | `volatile`        | <pre>object</pre><br/>(`null`)  | subscription information, used in [user join/leave notifications](/core/2/api/essentials/volatile-data) |
+| `cluster`        | <pre>boolean</pre><br/>(`true`)  | If true, realtime notifications will be triggered no matter which node receive a request |
```
Leodau, line 43:
```suggestion
| `cluster`        | <pre>boolean</pre><br/>(`true`)  | If true, realtime notifications will be triggered no matter which node receives a request |
```

Aschen:
> A couple issues:
> 
>     * you should mention what versions of kuzzle accept this option
> 
>     * this looks like a circular dependency in disguise; why do we need to release the support for this option before we release the version of kuzzle accepting it?

I'm still not sure about what Kuzzle version it's gonna be. It's an enhancement since the SDK is already exposed but I still feel like being able to use it with the HotelClerc features is a big improvement and it may deserve the new-feature tag IMHO, what do you think ?

About the circular dependency, I have no choice unfortunately because I cannot use the `query` method since I need to use the `realtime.subscribe` method with the `Room` mechanism.. (It would have been possible with this feature: https://github.com/kuzzleio/sdk-javascript/pull/489 :grin: )

But in this case it's not a big deal since this option is not recognized by Kuzzle for now.

scottinet:
What do you think about hiding the documentation (you can left it there, but comment it), so that once Kuzzle is released with that feature, you can uncomment the SDK documentation, adding the right Kuzzle version (and a link to the appropriate guide/API doc)?

I fear that documenting a feature that cannot be used (at the moment) is not a good idea.

Aschen:
@scottinet I'm OK with that.
And also maybe we should not document this feature at all in the SDK but only in the plugin/application documentation since the options is useless outside a plugin, what do you think?

Aschen:
@Leodau @scottinet 
I removed all the code related to the `cluster` option since this option is now added by the EmbeddedSDK only.
So this PR is just about few minor stuff now

src/core/security/Profile.js
```diff
@@ -5,7 +5,11 @@ class Profile {
    * @param {Object} data
    */
   constructor (kuzzle, _id = null, content = null) {
-    this._kuzzle = kuzzle;
+    Reflect.defineProperty(this, '_kuzzle', {
+      value: kuzzle,
+      writable: true
```
scottinet, line 10: ... we REALLY need to refactor these classes :expressionless:

src/core/security/Role.js
```diff
@@ -5,7 +5,11 @@ class Role {
    * @param {Object} data
    */
   constructor (kuzzle, _id = null, controllers = {}) {
-    this._kuzzle = kuzzle;
+    Reflect.defineProperty(this, '_kuzzle', {
+      value: kuzzle,
+      writable: true
```
scottinet, line 10: Why is this property writable?

PR #526, opened 2020-07-14
Add typescript definitions for Auth controller

Aschen:
## What does this PR do?

Adds Typescript support for the AuthController.

I took the descriptions from the documentation.

Build and pack the SDK: `npm run build && npm pack`

Then init a Node project and add this dependency:
```
{
  ...
  "dependencies": {
    "kuzzle-sdk": "file:kuzzle-sdk-7.2.0.tgz"
  }
}
```

Then create a `test.ts` file with the following and try to use the controller (just for the auto-completion)


src/Kuzzle.ts
```diff
@@ -259,14 +314,18 @@ class Kuzzle extends KuzzleEventEmitter {
...
         return false;
       }
       protectedEvent.lastEmitted = now;
     }
 
-    super.emit(eventName, ...payload);
+    return this._superEmit(eventName, ...payload);
+  }
+
+  _superEmit (eventName, ...payload) {
```
scottinet, line 327: What's the purpose of this function? :thinking: 

(same goes for  [_superAddListener()](https://github.com/kuzzleio/sdk-javascript/pull/526/files#diff-5a8206688414a52fe7efb0e1bef73b1cR421))

src/controllers/Auth.ts
```diff
@@ -0,0 +1,529 @@
...
+      this._authenticationToken = new Jwt(encodedJwt);
+    }
+    else {
+      throw new Error(`Invalid token argument: ${encodedJwt}`);
+    }
+  }
+
+  /**
+   * Do not add the token for the checkToken route, to avoid getting a token error when
+   * a developer simply wish to verify his token
```
scottinet, line 50: Use the gender-neutral [singular they](https://en.wikipedia.org/wiki/Singular_they):

```suggestion
   * a developer simply wishes to verify their token
```

src/controllers/Auth.ts
```diff
@@ -0,0 +1,529 @@
...
+    return this.query(request, {queuable: false, verb: 'POST'})
+      .then(response => {
+        try {
+          this._authenticationToken = new Jwt(response.result.jwt);
+
+          this.kuzzle.emit('loginAttempt', {success: true});
+        }
+        catch (err) {
+          return Promise.reject(err);
+        }
```
scottinet, line 385: Redundant try..catch: an exception thrown inside a promise automatically rejects it.

src/controllers/Auth.ts
```diff
@@ -0,0 +1,529 @@
...
+   * @see https://docs.kuzzle.io/sdk/js/7/controllers/auth/refresh-token
+   *
+   * @param options Additional options
+   *    - "queuable" If true, queues the request during downtime, until connected to Kuzzle again
+   *    - "expiresIn" Expiration duration
+   *
+   * @returns The refreshed token
+   */
+  refreshToken(
+    options: { queuable?: boolean, expiresIn?: number } = {}
```
scottinet, line 496: [expiresIn](https://docs.kuzzle.io/core/2/api/controllers/auth/refresh-token/) can be either a number or a string

```suggestion
    options: { queuable?: boolean, expiresIn?: number|string } = {}
```

src/controllers/Auth.ts
```diff
@@ -0,0 +1,529 @@
...
+   * @param strategy Name of the strategy to use
+   * @param credentials Credentials object for the strategy
+   * @param expiresIn Expiration time in ms library format. (e.g. "2h")
+   *
+   * @returns The encrypted JSON Web Token
+   */
+  login (
+    strategy: string,
+    credentials: JSONObject,
+    expiresIn: string = null
```
scottinet, line 367: [expiresIn](https://docs.kuzzle.io/core/2/api/controllers/auth/login/) can be a number. Also, it's optional:

```suggestion
    expiresIn?: string|number = null
```

(I _think_ that this works, but I'm fairly new to TS)

src/Kuzzle.ts
Aschen: To simplify the test.. It's was much more easier to just mock `_superEmit` than to mock the import of `KuzzleEventEmitter` (it didn't work after the TS conversion)

scottinet: You normally don't need to do that, you can directly mock the parent class (e.g. https://github.com/kuzzleio/kuzzle/blob/master/test/core/security/userRepository.test.js#L90 )

I won't block the PR for that, though.

Aschen: I got a `TypeError: Attempted to wrap emit which is already wrapped` when I try to wrap it that way but when I remove the line I got `expected 'emit' ...  to be sinon spy or spy call`

PR #527, opened 2020-07-16
Add typescript support for Document controller

Aschen:
## What does this PR do?

Adds TS support for the Document controller


src/controllers/Document.ts
```diff
@@ -0,0 +1,805 @@
...
+   * @param query Query to match
+   * @param options Additional options
+   *    - `queuable` If true, queues the request during downtime, until connected to Kuzzle again
+   *
+   * @returns The number of matching documents
+   */
+  count (
+    index: string,
+    collection: string,
+    body: JSONObject,
```
scottinet, line 30: The body is optional:

```suggestion
    body: JSONObject = null,
```

src/controllers/Document.ts
```diff
@@ -0,0 +1,805 @@
...
+   */
+  update (
+    index: string,
+    collection: string,
+    _id: string,
+    content: JSONObject,
+    options: {
+      queuable?: boolean,
+      refresh?: string,
+      retryOnConflict?: boolean,
```
scottinet, line 702:
```suggestion
      retryOnConflict?: number,
```

PR #529, opened 2020-07-22
Add typescript support for protocols

Aschen:
## What does this PR do?

Convert protocols classes to TS

### Other changes

 - `sslConnection` is deprecated for `ssl` and the value is `true` by default for port `443` and `7443`
 - remove the `error.internal` error who contain the original WS connect error and put the info in the message instead


src/protocols/abstract/Base.ts
```diff
@@ -1,20 +1,46 @@
...
     super();
 
     this._pendingRequests = new Map();
     this._host = host;
     this._name = name;
     const port = parseInt(options.port, 10);
     this._port = isNaN(port) ? 7512 : port;
-    this._ssl = typeof options.sslConnection === 'boolean' ? options.sslConnection : false;
+
+    if (typeof options.ssl === 'boolean') {
```
scottinet, line 32: (nitpicking) we should probably add an assertion check verifying if either ssl OR sslConnection is set, but not both

src/protocols/abstract/Base.ts
```diff
@@ -1,20 +1,50 @@
...
 
     this._pendingRequests = new Map();
     this._host = host;
     this._name = name;
     const port = parseInt(options.port, 10);
     this._port = isNaN(port) ? 7512 : port;
-    this._ssl = typeof options.sslConnection === 'boolean' ? options.sslConnection : false;
+
+    if (options.ssl !== undefined && options.sslConnection !== undefined) {
+      throw new Error('Both "ssl" and "sslConnection" options has been set. Use only "ssl".');
```
scottinet, line 33:
```suggestion
      throw new Error('Both "ssl" and "sslConnection" options are set. Use only "ssl".');
```

src/protocols/Http.ts
```diff
@@ -114,13 +152,13 @@ class HttpProtocol extends KuzzleAbstractProtocol {
...
       .then(() => {
-        this._routes = Object.assign(this._routes, this.customRoutes);
+        this._routes = Object.assign(this._routes, this._customRoutes);
 
         // Client is ready
         this.clientConnected();
       })
       .catch(err => {
-        const connectionError = new Error(`Unable to connect to kuzzle server at ${this.host}:${this.port}`);
+        const connectionError: any = new Error(`Unable to connect to kuzzle server at ${this.host}:${this.port}`);
```
xbill82, line 161: Shouldn't this be...
```suggestion
        const connectionError: Error = new Error(`Unable to connect to kuzzle server at ${this.host}:${this.port}`);
```

Aschen: I can't use the `Error` type because the next line we add the original internal error in the `error.internal` property

src/protocols/abstract/Base.ts
```diff
@@ -1,20 +1,50 @@
...
 
     this._pendingRequests = new Map();
     this._host = host;
     this._name = name;
     const port = parseInt(options.port, 10);
     this._port = isNaN(port) ? 7512 : port;
-    this._ssl = typeof options.sslConnection === 'boolean' ? options.sslConnection : false;
+
+    if (options.ssl !== undefined && options.sslConnection !== undefined) {
+      throw new Error('Both "ssl" and "sslConnection" options are set. Use only "ssl".');
```
xbill82, line 33: Why don't we just trigger a `console.warn` when `sslConnection` is set, saying it's going to be deprecated in the next version?

Aschen: IMHO we should do this for every deprecated feature used by users but also we don't want 1) flood the console 2) print messages in production so even if in this case the message won't be printed twice (only at instantiation), this is a bigger subject about how we warn users when they use deprecated features.

@rolljee work on the deprecation module on Kuzzle will be the first step, then we will have to think about how to integrate deprecation notifications in the SDK for developers

scottinet: Plus, this error is thrown if BOTH "ssl" and "sslConnection" are set, because in that case, and if they have different value, then the SDK cannot decide what to do. So the best course of action is to prevent both those options to be set at the same time.

xbill82: I remember VueJS showing deprecation warning in development builds and hiding them in production builds. But we don't have access to such a notion, do we?

Aschen: Maybe we could since we use the same tools as any other frontend build
In Node environment we could rely on NODE_ENV

PR #530, opened 2020-07-22
Add typescript support for search results

Aschen:
## What does this PR do?

Adds Typescript support for SearchResult classes


src/core/searchResult/Specifications.ts
```diff
@@ -1,6 +1,7 @@
-const { SearchResultBase } = require('./SearchResultBase');
+import { SearchResultBase } from './SearchResultBase';
+import { JSONObject } from '../../utils/interfaces';
 
-class SpecificationsSearchResult extends SearchResultBase {
+export class SpecificationsSearchResult extends SearchResultBase<JSONObject> {
```
Aschen, line 4: I know I should have created a proper `Specification` interface but I'm lazy to do it since I doubt someone is using this and we need a refactor of this module.
The `JSONObject` type works like a POJO

src/core/searchResult/SearchResultBase.ts
```diff
@@ -80,10 +73,18 @@ export class SearchResultBase<T> implements SearchResult<T> {
     Reflect.defineProperty(this, '_response', {
       value: response
     });
-
-    this._controller = request.controller;
-    this._searchAction = 'search';
-    this._scrollAction = 'scroll';
+    Reflect.defineProperty(this, '_controller', {
+      value: request.controller,
+      writable: true
```
xbill82, line 78: Why are these writable?

Aschen: Because in the `SpecificationSearchResult` class we overload the controller with `collection`

PR #537, opened 2020-08-10
Add typescript support realtime

Aschen:
## What does this PR do?

Convert realtime controller to TS

:warning: The whole SDK does not works with TS now since we need to convert _every_ exported class


src/controllers/Realtime.ts
```diff
@@ -0,0 +1,259 @@
...
+        return response.result;
+      });
+  }
+
+  /**
+   * Internal method.
+   *
+   * Should be called on network error or disconnection
+   */
+  disconnected () {
```
xbill82, line 202: 🤔 can't we make this private/protected, since it's an internal method? Oh, it probably needs to be called by the Protocol class so that might be a reason to leave it public.
Ok, so we have something similar to the "loose-coupling" problem we had in the core.
By the way, the naming of this method looks confusing to me. If it was a handler, I'd call it `onDisconnected()`, which tells that "what goes in this block gets executed when the connection is closed", which is OK, but I'd prefer the name of the method tells _what the method does_, i.e. "delete all the subscriptions of this instance of the SDK".
I posted [a similar comment on a PR in the core](https://github.com/kuzzleio/kuzzle/pull/1735#issuecomment-671427457), so I'd like to leave my 2 cents on the naming of this method (if it's not a breaking change)

* deleteAllSubscriptions()
* deleteAllRooms()

👋

src/utils/interfaces.ts
```diff
@@ -185,3 +185,136 @@ export interface DocumentHit extends Document {
...
   _score: number;
 }
+
+/**
+ * Enum for notification types
+ */
+export enum ENotificationType {
+  document = 'document',
+  user = 'user',
+  TokenExpired = 'TokenExpired'
```
xbill82, line 195: Why PascalCase?

Aschen: Because Kuzzle sent the event like this :(

src/controllers/Realtime.ts
Aschen: I take advantage of existing events and I renamed the methods to `saveSubscriptions` and `resubscribe`

PR #546, opened 2020-09-15
Remove "scroll" option from security:searchRoles doc

Yoann-Abbes:
## What does this PR do?

Remove "scroll" option from security:searchRoles doc

## Other change

Change script test name for admin console tests
depends on https://github.com/kuzzleio/kuzzle-admin-console/pull/829

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,7 +97,7 @@ jobs:
           npm ci
           npm install kuzzle-sdk-v7@$TRAVIS_BUILD_DIR
           cd test/e2e/run-test && npm ci && cd -
-          CYPRESS_RETRIES=5 npm run e2e
+          CYPRESS_RETRIES=5 npm run test:e2e
         fi
 
     - stage: Tests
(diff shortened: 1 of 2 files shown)


scottinet:
Shouldn't this PR target 7-dev?

Yoann-Abbes:
> Shouldn't this PR target 7-dev?

I thought I wouldnt be a problem to merge it directly on master to fix this error as soon as possible.
What do you think?

scottinet:
I don't see why this would be an emergency: the method isn't event documented.

Yoann-Abbes:
> I don't see why this would be an emergency: the method isn't event documented.

Base branch changed to 7-dev

scottinet:
Uh... so, I finally read the title, description and code of this PR. I thought this was about removing some code, not a documented option.

I apologize. And this PR, of course, should be merged on master. :expressionless:

scottinet (approved):

.travis.yml
```diff
@@ -97,7 +97,7 @@ jobs:
           npm ci
           npm install kuzzle-sdk-v7@$TRAVIS_BUILD_DIR
           cd test/e2e/run-test && npm ci && cd -
-          CYPRESS_RETRIES=5 npm run test:e2e
```
Aschen, line 100: End 2 end tests must be run with the `test:e2e` command. You should change the `package.json` if it's not the case

Aschen (approved):

.travis.yml
Yoann-Abbes: That means a PR on admin console repo and merge to master https://github.com/kuzzleio/kuzzle-admin-console/pull/829

PR #574, opened 2020-12-11
Adds a specific error when the request is too big to use the GET verb 

Aschen:
## What does this PR do?

HTTP headers request field have a size limite: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/431

When using the `document.mGet` method with the `get` verb, you can hit this limit who result in an 431 response with an empty body.

Now a specific error is trowed so developers can understand what is wrong

### DIFF
--- a/src/protocols/abstract/Base.ts
+++ b/src/protocols/abstract/Base.ts
@@ -122,7 +122,7 @@ export abstract class KuzzleAbstractProtocol extends KuzzleEventEmitter {
 Discarded request: ${JSON.stringify(request)}`));
     }
 
-    const stack = Error().stack;
+    let stack = Error().stack;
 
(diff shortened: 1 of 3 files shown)


Yoann-Abbes (approved):

src/protocols/abstract/Base.ts
```diff
@@ -122,7 +122,7 @@ export abstract class KuzzleAbstractProtocol extends KuzzleEventEmitter {
 Discarded request: ${JSON.stringify(request)}`));
     }
 
-    const stack = Error().stack;
+    let stack = Error().stack;
```
Aschen, line 125:
```suggestion
    const stack = Error().stack;
```

xbill82 (approved):

Issue #555, opened 2020-11-12
Add keep-alive mechanism to detect connection lost with realtime protocols


Aschen:
##  Feature Request

We need a keel-alive mechanism built with Kuzzle ping/pong capabilities with the WebSocket protocol to terminate a connection when the connection is lost.

See: https://www.npmjs.com/package/ws#how-to-detect-and-close-broken-connections

Yoann-Abbes:
Related: https://github.com/kuzzleio/sdk-javascript/issues/277

PR #586, opened 2021-01-14
Add keep-alive mechanism to detect connection lost with realtime protocols

Yoann-Abbes:
## What does this PR do?

Terminate a connection when the connection is lost by using the ping/pong mechanism
https://github.com/kuzzleio/sdk-javascript/issues/555

### Other changes

fix ms/security controller tests


doc/7/protocols/websocket/properties/index.md
```diff
@@ -14,6 +14,7 @@ order: 10
 | `connected`  | <pre>boolean</pre>  | Returns `true` if the socket is open |
 | `host`  | <pre>string</pre>  | Kuzzle server host |
 | `reconnectionDelay`  | <pre>number</pre>  | Number of milliseconds between reconnection attempts         |
+| `pingInterval`  | <pre>number</pre>  | Number of milliseconds between two pings         |
```
scottinet, line 17: same here

src/protocols/WebSocket.ts
```diff
@@ -83,8 +85,17 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
 
       this.client = new this.WebSocketClient(url, this.options);
 
+      this.client.onping = () => {
+        this.heartbeat(this.client);
+      };
+
+      this.client.onpong = () => {
+        return resolve();
+      };
```
scottinet, line 94: 1. you shouldn't resolve the promise in `onpong`: you're resolving the promise multiple times
2. your 'onping' method will only work if kuzzle does send PINGs, but that may very well not be the case, as Kuzzle can be configured to have passive checks (i.e. it will responds to PINGs with PONGs, but it won't send PING packets by itself), see here: https://github.com/kuzzleio/kuzzle/blob/master/.kuzzlerc.sample#L419
3. onping/onpong methods don't exist in the browser implementation of websockets, you'll have to create PING packets manually, and check if a received packet is a pong one (see https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#pings_and_pongs_the_heartbeat_of_websockets)

What you should do instead is to manually send PING packets to Kuzzle every `pingInterval` milliseconds, and then check received messages to verify if it's a PONG response, which resets the heartbeat timeout.

So you shouldn't have "onping"/"onpong" methods at all.

scottinet (changes requested):

doc/7/protocols/websocket/introduction/index.md
```diff
@@ -13,3 +13,11 @@ Inherits from: [KuzzleEventEmitter](/sdk/js/7/core-classes/kuzzle-event-emitter)
 The WebSocket protocol can be used by an instance of the SDK to communicate with your Kuzzle server.
 
 This protocol allows you to use all the features of Kuzzle, including [real-time notifications](/sdk/js/7/essentials/realtime-notifications).
+
+<SinceBadge version="2.10.0"/>
```
Aschen, line 17: We are not sure about the target release version (even if in that case we know pretty much I admit)
```suggestion
<SinceBadge version="change-me"/>
```

doc/7/protocols/websocket/introduction/index.md
```diff
@@ -13,3 +13,11 @@ Inherits from: [KuzzleEventEmitter](/sdk/js/7/core-classes/kuzzle-event-emitter)
...
+:::info
```
Aschen, line 18:
```suggestion
::: info
```

Aschen: I wish we could have a dedicated section so we can send links directly to people asking about this feature

src/protocols/WebSocket.ts
```diff
@@ -176,8 +175,13 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
       this.client.onmessage = payload => {
         const data = JSON.parse(payload.data || payload);
 
+        /**
+         * Since Kuzzle 2.10.0
+         * Corresponds to a custom pong response message
+         */
         if (data && data.p && data.p === '1') {
```
Aschen, line 182: The ping message should be a integer `{ "p": 1}`

Aschen (changes requested):

doc/7/protocols/websocket/constructor/index.md
```diff
@@ -30,8 +30,9 @@ WebSocket protocol connection options.
 | Property            | Type<br/>(default)               | Description                                                                                  |
 | ------------------- | -------------------------------- | -------------------------------------------------------------------------------------------- |
 | `autoReconnect`     | <pre>boolean</pre><br/>(`true`)  | Automatically reconnect to kuzzle after a `disconnected` event                               |
-| `port`              | <pre>number</pre><br/>(`7512`)   | Kuzzle server port                                                                           |
 | `headers`           | <pre>object</pre>(`{}`)          | Connection custom HTTP headers (e.g. origin, subprotocols, ...)<br/>**(Not supported by browsers)** |
+| `pingInterval` | <pre>number</pre><br/>(`30000`)   | Number of milliseconds between two pings                                         |
```
Aschen, line 34: I think `30s` is too high. On a mobile application you want to quickly know if you lost the connection so the SDK can start queuing requests.

The `2s` delay used in templates seems good to me.

doc/7/protocols/websocket/introduction/index.md
```diff
@@ -13,3 +13,11 @@ Inherits from: [KuzzleEventEmitter](/sdk/js/7/core-classes/kuzzle-event-emitter)
 The WebSocket protocol can be used by an instance of the SDK to communicate with your Kuzzle server.
 
 This protocol allows you to use all the features of Kuzzle, including [real-time notifications](/sdk/js/7/essentials/realtime-notifications).
+
+<SinceBadge version="change-me"/>
+::: info
```
Aschen, line 18: We need a dedicated section here IMHO so we could link this section directly to clients.
Proposal: `### Ping / Pong keep-alive`

src/protocols/WebSocket.ts
```diff
@@ -83,8 +86,36 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
...
       this.client.onopen = () => {
         this.clientConnected();
+        /**
+         * Send pings to the server
+        */ 
+        this.pingIntervalId = setInterval(() => {
+          this.ping();
+          this.pongTimeoutId = setTimeout(() => {
+            this.close();
+            return;
```
Aschen, line 116: Useless return

src/protocols/WebSocket.ts
```diff
@@ -148,8 +188,15 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
...
           this.emit('queryError', error, data);
         }
+        /**
+         * In case you're running a Kuzzle version under 2.10.0
+         * The response from a browser custom ping will be an error
+         * since it does not parse the client message properly.
+         * We need to clear this timeout at each message to keep 
+         * the connection alive if it's the case
+         */
+        clearTimeout(this.pongTimeoutId);
```
Aschen, line 198: Shouldn't this be in the error handling `else` bloc according to the comment?

Also I don't understand the message, if we receive a message from Kuzzle we don't care if it's an error or not. Since we received a message then we should clean the pong timer and consider the connection as alive.

It seems that this is the actual behavior but then I think your comment is wrong.

Aschen (changes requested):

src/protocols/WebSocket.ts
Yoann-Abbes: The comment was not accurate

doc/7/protocols/websocket/constructor/index.md
Yoann-Abbes: I didn't assume that the primary use was for a mobile application.
If this is the case, the option remains configurable.
On the other hand, if it's a mobile application, it means more network activity, and therefore more battery consumption.
I'll go for something like 10 seconds by default. 
Wdyt? @scottinet  @Aschen  ?

doc/7/protocols/websocket/constructor/index.md
```diff
@@ -30,8 +30,9 @@ WebSocket protocol connection options.
 | Property            | Type<br/>(default)               | Description                                                                                  |
 | ------------------- | -------------------------------- | -------------------------------------------------------------------------------------------- |
 | `autoReconnect`     | <pre>boolean</pre><br/>(`true`)  | Automatically reconnect to kuzzle after a `disconnected` event                               |
-| `port`              | <pre>number</pre><br/>(`7512`)   | Kuzzle server port                                                                           |
 | `headers`           | <pre>object</pre>(`{}`)          | Connection custom HTTP headers (e.g. origin, subprotocols, ...)<br/>**(Not supported by browsers)** |
+| `pingInterval` | <pre>number</pre><br/>(`10000`)   | Number of milliseconds between two pings                                         |
```
Aschen, line 34: As I said, 10 seconds seems to be a lot. In 10 seconds an user can do many interactions with a mobile or web app before the application realize that there is no longer connection with the backend and switch to offline mode.

2 seconds seems to be small enough to have more reactivity.

(@scottinet @Njuelle any thought on this?)

Aschen (changes requested):
One last comment / discussion about the default `pingInterval` and then LGTM :+1:

doc/7/protocols/websocket/constructor/index.md
scottinet: dkdc :expressionless:

doc/7/protocols/websocket/constructor/index.md
Njuelle: I agree, 10 seconds in an app that is currently in use is too long.
2 seconds is ok for me too

src/protocols/WebSocket.ts
```diff
@@ -105,20 +105,24 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
...
       this.client.onopen = () => {
-        this.clientConnected();
+        this.clientConnected().then(() => {
+          this.pingIntervalId = setInterval(() => {
+            this.ping();
+            this.pongTimeoutId = setTimeout(() => {
+              this.client.close();
+            }, this._pongTimeout);
+          }, this._pingInterval);
+        });
```
Aschen: Nitpicking (readability): I struggle a little bit to read this part

Aschen (approved):

src/protocols/WebSocket.ts
Yoann-Abbes: read it again please I've changed the logic

test/protocol/WebSocket.test.js
```diff
@@ -332,6 +344,30 @@ describe('WebSocket networking module', () => {
...
+    const
+      cb = sinon.stub(),
+      cb2 = sinon.stub();
+
+    websocket.on('discarded', cb);
+    websocket.on('queryError', (error, data) => {
+      cb2(error, data);
+    });
+    
+    websocket.connect();
```
alexandrebouthinon, line 357: Missing await?

test/protocol/WebSocket.test.js
```diff
@@ -99,6 +99,16 @@ describe('WebSocket networking module', () => {
     should(websocket.stopRetryingToConnect).be.false();
   });
 
+  it('should initialize a ping interval when the connection is established', () => {
+    const setInterval = sinon.stub(clock, 'setInterval');
+
+    websocket.connect();
```
alexandrebouthinon, line 105: Missing await?

alexandrebouthinon (changes requested):

src/protocols/WebSocket.ts
```diff
@@ -137,6 +171,15 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
       this.client.onmessage = payload => {
         const data = JSON.parse(payload.data || payload);
 
+        /**
+         * Since Kuzzle 2.10.0
+         * Corresponds to a custom pong response message
+         */
+        if (data && data.p && data.p === 2) {
```
scottinet, line 178: you need to make sure that this is a PONG payload, by checking that there are no other properties than the `p` one (imagine a custom API route returning a response containing a numeral "p" property)

src/protocols/abstract/Realtime.ts
```diff
@@ -5,15 +5,21 @@ import { KuzzleAbstractProtocol } from './Base';
...
   protected retrying: boolean;
+  protected pongTimeoutId: ReturnType<typeof setTimeout>;
+  protected pingIntervalId: ReturnType<typeof setInterval>;
 
   constructor (host, options: any = {}, name: string) {
     super(host, options, name);
 
     this._autoReconnect = typeof options.autoReconnect === 'boolean' ? options.autoReconnect : true;
     this._reconnectionDelay = typeof options.reconnectionDelay === 'number' ? options.reconnectionDelay : 1000;
+    this._pingInterval = typeof options.pingInterval === 'number' ? options.pingInterval : 2000;
```
scottinet, line 21: documentation and comments still state that the default value is 10s

scottinet (changes requested):

test/protocol/WebSocket.test.js
Yoann-Abbes: No

test/protocol/WebSocket.test.js
Yoann-Abbes: same as above

alexandrebouthinon (approved):

PR #602, opened 2021-02-15
Fix re-subscription

Aschen:
## What does this PR do?

When resubscribing after a `reconnected` event, the SDK use the original subscription `request` to re-subscribe.

The `Kuzzle.query` method mutate request and inject the JWT inside. Meaning that the request saved for the re-subscription already has a JWT inside.

This can be an issue because developers rely on the `kuzzle.jwt` property of the SDK to know if 1) the token is still valid 2) there are connection information.

Now the `request` and `options` parameter of the query method are cloned before modification.  (:warning: to save resources it's not a deep clone so only the top level properties are duplicated. Nested objects will not be cloned but only referenced)

Run a Kuzzle stack and then create this user and update anonymous role:
```
kourou security:createUser '{                                                             
  content: {
    profileIds: ["default"]
  },
  credentials: {
    local: {
      username: "test-admin",
      password: "password"
    },
  }
}'

kourou security:updateRole '{                                                             
  controllers: {
    "realtime": {
      actions: {
        "subscribe": false
      }
    },
    "*": {
      actions: {
        "*": true
      }
    }
  }
}'
```

Then run this script

<details><summary>resubscribe.js</summary>

```js
const { Kuzzle, WebSocket } = require('./index');

const kuzzle = new Kuzzle(new WebSocket('localhost'));

async function subscribeToNotifications () {
  console.log('subscribeToNotifications');
  await kuzzle.realtime.subscribe('test', 'test', {}, () => {
    console.log('Notification received');
  });
  console.log('Subscribed');
}

// authenticate and subscribe after a "tokenExpired" event
kuzzle.on('tokenExpired', async () => {
  await authenticateAndSubscribe();
});

kuzzle.on('reconnected', async () => {
  await authenticateAndSubscribe();
})

// authenticate and then subscribe
// re-try every 1 second until it works
async function authenticateAndSubscribe () {
  console.log('authenticateAndSubscribe()');
  try {
    const { valid } = await kuzzle.auth.checkToken();

    if (! valid) {
      await kuzzle.auth.login('local', { username: 'test-admin', password: 'password' }, '5s');
      console.log('Authenticated');
    }
    else {
      console.log('Already authenticated');
    }

    await subscribeToNotifications();
  }
  catch (error) {
    console.log('authenticateAndSubscribe ', error);
    // Wrong or revoked credentials
    if (error.id === 'plugin.strategy.missing_user') {
      throw error;
    }

    // An error occured (network, etc), try again in 1 sec
    setTimeout(async () => {
      try {
        await authenticateAndSubscribe();
      }
      catch (error) {
        // if we end up there it means that we don't have correct credentials anymore
      }
    }, 1000);
  }
}

async function refreshToken () {
  setInterval(async () => {
    try {
      await kuzzle.auth.refreshToken({ expiresIn: '5s' });
      console.log('Refreshed')
    }
    catch (error) {
      console.log('Cannot refresh token: ', error);
    }
  }, 4000);
}

async function run () {
  await kuzzle.connect();
  await authenticateAndSubscribe();

  await refreshToken();
}

run();
```

</details>

<details><summary>Logs</summary>

```
➜  sdk-javascript git:(fix-resubscription) ✗ node -r ts-node/register test.js
authenticateAndSubscribe()
Authenticated
subscribeToNotifications
Subscribed
Cannot refresh token:  Error: Unable to execute request: not connected to a Kuzzle server.
Discarded request: {"action":"refreshToken","expiresIn":"5s","controller":"auth","requestId":"02248d26-4cd9-4661-b5bf-628c95b30063","volatile":{"sdkInstanceId":"f346463c-2674-4753-8ab6-4ad8054157cd","sdkName":"js@7.5.3"},"jwt":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJfaWQiOiI4MjcwZTIxMS1kM2E5LTQzM2UtYTgzMS02ODkzZDUwMzU3ZDgiLCJpYXQiOjE2MTMzOTU1NTMsImV4cCI6MTYxMzM5NTU1OH0.cSfYw97kY_uah623nCRUnpOjXw6Zp1eJtEb3lj7zfwk"}
    at WebSocketProtocol.query (/home/aschen/projets/kuzzleio/sdk-javascript/src/protocols/abstract/Base.ts:121:29)
    at Kuzzle.query (/home/aschen/projets/kuzzleio/sdk-javascript/src/Kuzzle.ts:589:26)
    at AuthController.query (/home/aschen/projets/kuzzleio/sdk-javascript/src/controllers/Base.ts:41:25)
    at AuthController.refreshToken (/home/aschen/projets/kuzzleio/sdk-javascript/src/controllers/Auth.ts:534:17)
    at Timeout._onTimeout (/home/aschen/projets/kuzzleio/sdk-javascript/test.js:57:25)
    at listOnTimeout (internal/timers.js:549:17)
    at processTimers (internal/timers.js:492:7)
Cannot refresh token:  Error: Unable to execute request: not connected to a Kuzzle server.
Discarded request: {"action":"refreshToken","expiresIn":"5s","controller":"auth","requestId":"ac2f20c0-5bbf-4469-8f91-e916db0d10e9","volatile":{"sdkInstanceId":"f346463c-2674-4753-8ab6-4ad8054157cd","sdkName":"js@7.5.3"},"jwt":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJfaWQiOiI4MjcwZTIxMS1kM2E5LTQzM2UtYTgzMS02ODkzZDUwMzU3ZDgiLCJpYXQiOjE2MTMzOTU1NTMsImV4cCI6MTYxMzM5NTU1OH0.cSfYw97kY_uah623nCRUnpOjXw6Zp1eJtEb3lj7zfwk"}
    at WebSocketProtocol.query (/home/aschen/projets/kuzzleio/sdk-javascript/src/protocols/abstract/Base.ts:121:29)
    at Kuzzle.query (/home/aschen/projets/kuzzleio/sdk-javascript/src/Kuzzle.ts:589:26)
    at AuthController.query (/home/aschen/projets/kuzzleio/sdk-javascript/src/controllers/Base.ts:41:25)
    at AuthController.refreshToken (/home/aschen/projets/kuzzleio/sdk-javascript/src/controllers/Auth.ts:534:17)
    at Timeout._onTimeout (/home/aschen/projets/kuzzleio/sdk-javascript/test.js:57:25)
    at listOnTimeout (internal/timers.js:549:17)
    at processTimers (internal/timers.js:492:7)
authenticateAndSubscribe()
Authenticated
subscribeToNotifications
Subscribed
Refreshed
Refreshed
```

</details>

Wait 2 sec and then stop Kuzzle:
 - refreshToken request should be queued
 - networkError should appear

Restart Kuzzle:
 - the automatic re-subscription will fail because the token is expired
 - the script should re-authenticate
 - only 1 subscription is registered (try `kourou realtime:publish test test '{ toto: 42 }' --username test-admin --password password` )

### Other changes

 - remove the `checkToken` when the SDK reconnect because it breaks the workflow on `tokenExpired` events

### DIFF
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -13,6 +13,7 @@ import { MemoryStorageController } from './controllers/MemoryStorage';
 
 import { uuidv4 } from './utils/uuidv4';
 import { proxify } from './utils/proxify';
+import { clone } from './utils/clone';
 import { JSONObject } from './types';
 import { RequestPayload } from './types/RequestPayload';
 import { ResponsePayload } from './types/ResponsePayload';
@@ -420,7 +421,9 @@ export class Kuzzle extends KuzzleEventEmitter {
       this.startQueuing();
     }
 
-    this.protocol.addListener('queryError', (err, query) => this.emit('queryError', err, query));
+    this.protocol.addListener('queryError', ({ error, request }) => {
+      this.emit('queryError', { error, request });
+    });
 
     this.protocol.addListener('tokenExpired', () => this.tokenExpired());
 
@@ -454,20 +457,6 @@ export class Kuzzle extends KuzzleEventEmitter {
 
       if (this.autoReplay) {
         this.playQueue();
-      }
-
-      if (this.auth.authenticationToken) {
-        return this.auth.checkToken()
-          .then(res => {
-            // shouldn't obtain an error but let's invalidate the token anyway
-            if (!res.valid) {
-              this.auth.authenticationToken = null;
-            }
-          })
-          .catch(() => {
-            this.auth.authenticationToken = null;
-          })
-          .then(() => this.emit('reconnected'));
       }
 
       this.emit('reconnected');
@@ -526,7 +515,10 @@ export class Kuzzle extends KuzzleEventEmitter {
    * @param request
    * @param options - Optional arguments
    */
-  query (request: RequestPayload = {}, options: JSONObject = {}): Promise<ResponsePayload> {
+  query (req: RequestPayload = {}, opts: JSONObject = {}): Promise<ResponsePayload> {
+    const request = clone(req);
+    const options = clone(opts);
+
     if (typeof request !== 'object' || Array.isArray(request)) {
       throw new Error(`Kuzzle.query: Invalid request: ${JSON.stringify(request)}`);
     }
@@ -590,7 +582,7 @@ export class Kuzzle extends KuzzleEventEmitter {
         });
       }
 
-      this.emit('discarded', {request});
+      this.emit('discarded', { request });
       return Promise.reject(new Error(`Unable to execute request: not connected to a Kuzzle server.
 Discarded request: ${JSON.stringify(request)}`));
     }
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -523,8 +523,8 @@ export class Kuzzle extends KuzzleEventEmitter {
       throw new Error(`Kuzzle.query: Invalid "options" argument: ${JSON.stringify(opts)}`);
     }
 
-    const request = { ...req };
-    const options = { ...opts };
+    const request = JSON.parse(JSON.stringify(req));
+    const options = JSON.parse(JSON.stringify(opts));
 
     if (!request.requestId) {
       request.requestId = uuidv4();
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -13,7 +13,6 @@ import { MemoryStorageController } from './controllers/MemoryStorage';
 
 import { uuidv4 } from './utils/uuidv4';
 import { proxify } from './utils/proxify';
-import { clone } from './utils/clone';
 import { JSONObject } from './types';
 import { RequestPayload } from './types/RequestPayload';
 import { ResponsePayload } from './types/ResponsePayload';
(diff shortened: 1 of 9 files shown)


src/Kuzzle.ts
```diff
@@ -526,15 +514,18 @@ export class Kuzzle extends KuzzleEventEmitter {
    * @param request
    * @param options - Optional arguments
```
scottinet, line 515:
```suggestion
   * @param req
   * @param opts - Optional arguments
```

src/Kuzzle.ts
```diff
@@ -526,15 +514,18 @@ export class Kuzzle extends KuzzleEventEmitter {
...
     }
 
-    if (typeof options !== 'object' || Array.isArray(options)) {
-      throw new Error(`Kuzzle.query: Invalid "options" argument: ${JSON.stringify(options)}`);
+    if (typeof opts !== 'object' || Array.isArray(opts)) {
+      throw new Error(`Kuzzle.query: Invalid "options" argument: ${JSON.stringify(opts)}`);
     }
 
+    const request = { ...req };
+    const options = { ...opts };
```
scottinet, line 527: (nitpicking) or you could make a deep-copy, as it is safer in case users modify the referenced objects before they are submitted again, if they get queued

```suggestion
    const request = JSON.parse(JSON.stringify(req));
    const options = JSON.parse(JSON.stringify(opts));
```

scottinet (approved):
A couple nitpickings

src/Kuzzle.ts
Aschen: That's what I was doing at first but then I thought about the performance drop and I sticked to the light copy. 
The query method is only mutating first level properties so IMHO it's enough and if users mutate there own payloads I think it's not our responsibility anymore

scottinet: About performances, I believe the JSON.stringify/JSON.parse method performs at roughly a couple millions instructions per second, so we should be quite safe there.

PR #613, opened 2021-03-09
Implement index:stats

### LINKED ISSUE #608: Implement API action index:stats
Implement API action index:stats

Leodau:
## What does this PR do?
This **PR** implements the recently added [index:stats](https://docs.kuzzle.io/core/2/api/controllers/index/stats/) _(@ Kuzzle Backend v2.10.0)_
So it can be used directly, instead of having to use **query** to access it.

#### Todo:

- [x] Implementation
- [x] Unit Testing
- [x] Doc + Code Snippet

1. Launch a Kuzzle stack.
2. Directly use `sdk.index.stats()`
3. Verify that you get stats that make sense.

Closes #608 

> :information_source: It seems that **Codecov** is not taking the latest commit? :facepalm:

### DIFF
--- a/doc/7/controllers/index/stats/snippets/stats.js
+++ b/doc/7/controllers/index/stats/snippets/stats.js
@@ -2,18 +2,20 @@ const stats = await kuzzle.index.stats();
 
 console.log(JSON.stringify(stats));
 /*
-  indexes: [
-    {
-      name: 'nyc-open-data',
-      size: 42,
-      collections: [
-        {
-          name: 'yellow-taxi',
-          documentCount: 42,
-          size: 42,
-        }
-      ]
-    }
-  ],
-  size: 42
+  {
+    "indexes":[
+      {
+        "name":"nyc-open-data",
+        "size":42,
+        "collections":[
+          {
+            "name":"yellow-taxi",
+            "documentCount":42,
+            "size":42
+          }
+        ]
+      }
+    ],
+    "size":42
+  }
 */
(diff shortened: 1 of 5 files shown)


scottinet (approved):

doc/7/controllers/index/stats/snippets/stats.js
```diff
@@ -0,0 +1,3 @@
+const stats = await kuzzle.index.stats();
+
+console.log(JSON.stringify(stats));
```
Aschen, line 3: Can you add the JSON structure in comment please?

So user know how to manipulate the result then

Aschen (changes requested):

src/controllers/Index.ts
```diff
@@ -99,4 +99,19 @@ export class IndexController extends BaseController {
...
+
+  /**
+   * Returns detailed storage usage statistics.
+   *
+   * @see https://docs.kuzzle.io/sdk/js/7/controllers/index/stats/
+   *
+   * @param options Additional options
+   *    - `queuable` If true, queues the request during downtime, until connected to Kuzzle again
+   */
+  stats (options: { queuable?: boolean } = {}): Promise<any> {
```
Aschen, line 111:
```suggestion
  stats (options: { queuable?: boolean } = {}): Promise<JSONObject> {
```

Aschen (approved):

src/controllers/Index.ts
Leodau: Oh good catch

alexandrebouthinon (commented):
To fix unit tests, you need to add this line at the top of `src/controllers/Index.ts`:
```js
import { JSONObject } from '../types';
```

alexandrebouthinon (approved):

PR #614, opened 2021-03-10
Print deprecation warning from Kuzzle

### LINKED ISSUE #593: Print deprecation warning from Kuzzle
## Feature Description

Deprecation warnings can be added to the response to inform the developer when he is using a deprecated feature.

IMHO those warnings should be printed to the developer when he is developing his application:
 - warnings should not be printed in production
 - warnings printing should be configurable

:warning: We need a small workshop first to discuss about this

See:
 - https://docs.kuzzle.io/core/2/framework/classes/kuzzle-request/add-deprecation/#_Classes_KuzzleRequest_addDeprecation

MathieuVeber:
Closes #593

## What does this PR do?
From now on, when developing with the SDK, you will see warning when using a deprecated action. You can disable this behavior by setting to false the deprecationWarning option on the kuzzle sdk object

Use a deprecated action and observe the warning(s)

### Other changes
Updated doc and added tests

### DIFF
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -528,7 +528,7 @@ export class Kuzzle extends KuzzleEventEmitter {
    * @param req
    * @param opts - Optional arguments
    */
-  query (req: RequestPayload = {}, opts: JSONObject = {}): Promise<ResponsePayload> {
+  async query (req: RequestPayload = {}, opts: JSONObject = {}): Promise<ResponsePayload> {
     if (typeof req !== 'object' || Array.isArray(req)) {
       throw new Error(`Kuzzle.query: Invalid request: ${JSON.stringify(req)}`);
     }
@@ -600,8 +600,9 @@ export class Kuzzle extends KuzzleEventEmitter {
 Discarded request: ${JSON.stringify(request)}`));
     }
 
-    return this.protocol.query(request, options)
-      .then((response) => this.deprecationHandler.logDeprecation(response));
+    const response = await this.protocol.query(request, options);
+
+    return this.deprecationHandler.logDeprecation(response);
   }
 
   /**
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -601,7 +601,7 @@ Discarded request: ${JSON.stringify(request)}`));
     }
 
     return this.protocol.query(request, options)
-      .then((response: ResponsePayload) => this.deprecationHandler.logDeprecation(response))
+      .then((response: ResponsePayload) => this.deprecationHandler.logDeprecation(response));
   }
 
   /**
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -528,7 +528,7 @@ export class Kuzzle extends KuzzleEventEmitter {
    * @param req
    * @param opts - Optional arguments
    */
-  async query (req: RequestPayload = {}, opts: JSONObject = {}): Promise<ResponsePayload> {
+  query (req: RequestPayload = {}, opts: JSONObject = {}): Promise<ResponsePayload> {
     if (typeof req !== 'object' || Array.isArray(req)) {
       throw new Error(`Kuzzle.query: Invalid request: ${JSON.stringify(req)}`);
     }
@@ -600,9 +600,8 @@ export class Kuzzle extends KuzzleEventEmitter {
 Discarded request: ${JSON.stringify(request)}`));
     }
 
-    const response = await this.protocol.query(request, options) as ResponsePayload;
-
-    return this.deprecationHandler.logDeprecation(response);
+    return this.protocol.query(request, options)
+      .then((response: ResponsePayload) => this.deprecationHandler.logDeprecation(response))
   }
 
   /**
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -600,7 +600,7 @@ export class Kuzzle extends KuzzleEventEmitter {
 Discarded request: ${JSON.stringify(request)}`));
     }
 
-    const response = await this.protocol.query(request, options);
+    const response = await this.protocol.query(request, options) as ResponsePayload;
 
     return this.deprecationHandler.logDeprecation(response);
   }
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -11,6 +11,7 @@ import { ServerController } from './controllers/Server';
 import { SecurityController } from './controllers/Security';
 import { MemoryStorageController } from './controllers/MemoryStorage';
 
+import { Deprecation } from './utils/Deprecation';
 import { uuidv4 } from './utils/uuidv4';
 import { proxify } from './utils/proxify';
 import { JSONObject } from './types';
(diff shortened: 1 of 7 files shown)


src/Kuzzle.ts
```diff
@@ -164,6 +169,11 @@ export class Kuzzle extends KuzzleEventEmitter {
        * If set to `auto`, the `autoQueue` and `autoReplay` are also set to `true`
        */
       offlineMode?: 'auto';
+      /**
+       * Show deprecation warning in development mode (hidden either way in production)
+       * Default: `true`
+       */
+      deprecationWarning?: boolean;
```
jenow, line 176: Shouldn't you put a default value here?

MathieuVeber: Well the default value doesn't seems to be defined here when looking at the others parameters but rather just after (from line 202)

src/utils/Deprecation.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  private _deprecationWarning: boolean;
+
+  constructor (deprecationWarning: boolean) {
+    this._deprecationWarning = deprecationWarning;
+  }
+
+  logDeprecation (response) {
+    if (this._deprecationWarning && response.deprecations?.length) {
+      response.deprecations.forEach(deprecation => {
```
Aschen, line 11: Usually we prefer the `for (const .. of )` loop:
 - `forEach` take a function so if you return inside the loop then you don't exit the main function
 - if you await a promise inside the function or return a promise, it's not awaited before starting calling the function on the next item
 
 Because of those two reason it's better to always have `for (const .. of ` loop for the sack of consistency

Aschen (changes requested):

Aschen:
I don't see how the deprecation are hidden in production by default?

src/utils/Deprecation.ts
MathieuVeber: Good to know!

MathieuVeber:
> I don't see how the deprecation are hidden in production by default?

If I understood well, usages of the console once the sdk is build are disabled, also it seems like Kuzzle server doesn't send deprecations info when turned in mode production.

But I might not be the best to answer that @jenow @Yoann-Abbes

src/controllers/Base.ts
```diff
@@ -1,9 +1,10 @@
+import { Kuzzle } from 'src/Kuzzle';
```
Aschen, line 1:
```suggestion
import { Kuzzle } from '../Kuzzle';
```

src/utils/Deprecation.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  /**
+   * Warn the developer that he is using a deprecated action
+   * 
+   * @param response Result of a query to the API
+   * 
+   * @returns Same as response param, just like a middleware
+   */
+  logDeprecation (response) {
```
Aschen, line 16: Missing type here. It should be a `ResponsePayload`

Aschen, line 16: Missing type here. IMHO it should be a `ResponsePayload`

src/utils/Deprecation.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (this._deprecationWarning && response.deprecations && response.deprecations.length) {
+      for (let index = 0; index < response.deprecations.length; index++) {
```
Aschen, line 18: Prefer `for (const ... of)` loop:
 - readability: no need to access elements by the array index
 - security: no out of range errors

src/utils/Deprecation.ts
```diff
@@ -0,0 +1,25 @@
...
+        // eslint-disable-next-line no-console
+        console.warn(response.deprecations[index].message);        
+      }
+    }
+    return response;
+  }
+}
```
Aschen, line 25: Missing new line

Aschen (changes requested):
Also you should ensure that the `Kuzzle` class is calling the deprecation util with an unit test

MathieuVeber:
> I don't see how the deprecation are hidden in production by default?

Finally, using process.env.NODE_ENV was the solution to this issue

src/Kuzzle.ts
```diff
@@ -528,7 +528,7 @@ export class Kuzzle extends KuzzleEventEmitter {
    * @param req
    * @param opts - Optional arguments
    */
-  query (req: RequestPayload = {}, opts: JSONObject = {}): Promise<ResponsePayload> {
+  async query (req: RequestPayload = {}, opts: JSONObject = {}): Promise<ResponsePayload> {
```
Aschen: You cannot use `async` in the SDK because then you need to transpile it for the browser and it will increase it size quite a lot

Aschen (changes requested):

src/Kuzzle.ts
MathieuVeber: But there not any async anymore...! See my lasts [commit](https://github.com/kuzzleio/sdk-javascript/pull/614/commits/4ed9f7338f351422b896aa2b8564ecc36310d18e#diff-fea3a80abb8ad2b947af003f0230a05d08929cbbeaa5e73d8a1f84e857b4755aR531)

src/types/ResponsePayload.ts
```diff
@@ -31,10 +31,27 @@ export type ResponsePayload = {
...
+    /**
+     * Deprecation description
+     */
+    message: string;
+
+    /**
+     * Deprecated since this version
+     */
+    version: string;
+  }[];
```
Aschen, line 48: Prefer the more readable `Array<{...}>` notation
```suggestion
  deprecations?: Array>{

    /**
     * Deprecation description
     */
    message: string;

    /**
     * Deprecated since this version
     */
    version: string;
  }>;
```

src/utils/Deprecation.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * Warn the developer that he is using a deprecated action (disabled if NODE_ENV=production)
+   * 
+   * @param response Result of a query to the API
+   * 
+   * @returns Same as response param, just like a middleware
+   */
+  logDeprecation (response: ResponsePayload) {
+    if ( process.env.NODE_ENV !== 'production'
```
Aschen, line 19: Reading the environment is quite expensive. Here you are gonna make an expensive syscall every time a request is sent to Kuzzle.

https://www.dynatrace.com/news/blog/the-drastic-effects-of-omitting-node-env-in-your-express-js-applications/

test/kuzzle/query.test.js
```diff
@@ -242,5 +242,14 @@ describe('Kuzzle query management', () => {
 
       should(request).be.eql({ controller: 'server', action: 'now' });
     });
+
+    it('should call logDeprecation with the response', async () => {
+      kuzzle.deprecationHandler.logDeprecation = sinon.stub().returns(response);
+      await kuzzle.query(query);
+
+      should(kuzzle.deprecationHandler.logDeprecation)
+        .be.calledOnce()
```
Aschen, line 251: Nitpicking: this line is useless since you check that the method is called with the response

Aschen (changes requested):

Aschen (approved):

test/kuzzle/query.test.js
MathieuVeber: But I make myself sure that the function is called only once, I feel like I'm also nitpicking!

PR #621, opened 2021-03-26
Secure WebSocket connection with cookie authentication

Shiranuit:
## What does this PR do ?

This PR adds the support of cookie when using the websocket protocol. To do that, when Kuzzle `cookieAuth` option is `true`, Kuzzle will call the method `enableCookieSupport` from the given protocol. When called, this method will throw if outside the browser or if cookie are not supported by the protocol, otherwise this will change how the protocol behave.

For the HTTP protocol it's simple, when enableCookieSupport is called, the protocol will be changing if the request are made with `withCredentials` set to `true` or `false` depending if he should be able to receive cookies.

For the Websocket protocol, this is a bit more complex, when enableCookieSupport is called, the protocol, will be creating a instance of the HTTP Protocol, with the same option (host, port, ssl, ...) as the websocket protocol, after that, when a request `auth:login`, `auth:logout` or `auth:refreshToken` is made, the protocol will use the HTTP Protocol instead of the websocket client to make the request. 
`[ex: auth:login request is made -> websocket closes the connection -> then send the request with the http protocol -> when a response is received it reopens the connection -> then resolve the request]`

renew auth

## TODO:

- [x] HTTP integration in Websocket Protcol
- [x] Unit Tests
- [x] Documentation
- [x] Add pictures to explain the changes in behaviour


src/protocols/Http.ts
```diff
@@ -173,12 +173,23 @@ export default class HttpProtocol extends KuzzleAbstractProtocol {
...
+  }
+
+  /**
+   * Preprocess and validate the request before sending
    *
    * @param {Object} data
    * @returns {Promise<any>}
    */
-  send (request: RequestPayload, options: JSONObject = {}) {
+  validateRequest (request: RequestPayload, options: JSONObject = {}) {
```
Aschen, line 192: A better naming could be `formatRequest`

src/protocols/Http.ts
```diff
@@ -330,6 +359,9 @@ export default class HttpProtocol extends KuzzleAbstractProtocol {
 
       xhr.open(method, url);
 
+      // Authorize the reception of cookies
+      xhr.withCredentials = this.cookieSupport;
```
Aschen, line 363: Shouldn't this be set only when cookie auth is activated?

src/protocols/WebSocket.ts
```diff
@@ -209,15 +211,58 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
...
+
+      if (verifiedRequest) {
+        if (this.client) {
+          this.client.close();
+        }
+        this.client = null;
+        this.clientDisconnected(); // Simulate a disconnection, this will enable offline queue and trigger realtime subscriptions backup
+
+        this._httpProtocol._sendHttpRequest(verifiedRequest.method, verifiedRequest.url, verifiedRequest.payload)
+          .then(async (response) => {
```
Aschen, line 256: You cannot use async/await in the SDK since it increase the build size

Aschen (changes requested):

src/protocols/Http.ts
Shiranuit: This is the case when `cookieSupport` is enabled in the protocol, the cookieSupport of a protocol is only enabled when `cookieAuth` option is `true` in the Kuzzle constructor

src/protocols/WebSocket.ts
Shiranuit: That seems weird, but ok I will change it

doc/7/protocols/websocket/introduction/index.md
```diff
@@ -23,3 +23,23 @@ This application-level PING has been especially added for web browsers, which do
 When run in a browser, our Javascript SDK uses that feature for its keep-alive mechanism: a message will periodically be sent to Kuzzle in the form `"{"p":1}"` through websocket.
 That message will call a response from Kuzzle in the form `"{"p":2}"` for the SDK to keep the connection alive.
 
+### Cookie Authentication
+
+Kuzzle does support cookie authentication, meaning that when you're making a request, you can instruct Kuzzle to store the token securely inside a cookie in your browser, instead of returning it in the response.
```
Aschen, line 28: Can we have the schema here for more clarity?

src/protocols/Http.ts
```diff
@@ -281,9 +292,27 @@ export default class HttpProtocol extends KuzzleAbstractProtocol {
...
+   * Sends a payload to the connected server
+   *
+   * @param {Object} data
+   * @returns {Promise<any>}
+   */
+  send (request: RequestPayload, options: JSONObject = {}) {
+    const verifiedRequest = this.formatRequest(request, options);
+
+    if (verifiedRequest) {
+      this._sendHttpRequest(verifiedRequest.method, verifiedRequest.url, verifiedRequest.payload)
```
Aschen, line 312: Nitpicking/boyscout: the `sendHttpRequest` could be refactored to take only the `verifiedRequest` object

src/protocols/WebSocket.ts
```diff
@@ -209,15 +211,61 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
...
    */
-  send (request: RequestPayload) {
-    if (this.client && this.client.readyState === this.client.OPEN) {
-      this.client.send(JSON.stringify(request));
+  send (request: RequestPayload, options: JSONObject = {}) {
+    if (!this.client || this.client.readyState !== this.client.OPEN) {
+      return;
+    }
+
+    if ( this.cookieSupport
```
Aschen, line 240: There is too many useless level of indentation starting here IMHO, you should rather use guard and try to put the code outside of the `if` blocks for readability

src/protocols/WebSocket.ts
```diff
@@ -209,15 +211,61 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
...
+        || request.action === 'refreshToken')
+    ) {
+      const verifiedRequest = this._httpProtocol.formatRequest(request, options);
+
+      if (verifiedRequest) {
+        if (this.client) {
+          this.client.close();
+        }
+        this.client = null;
+        this.clientDisconnected(); // Simulate a disconnection, this will enable offline queue and trigger realtime subscriptions backup
```
Aschen, line 253: Does it enable the queuing of requests even if those options were not activated on Kuzzle constructor?

src/protocols/WebSocket.ts
```diff
@@ -209,15 +211,61 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
...
+
+        this._httpProtocol._sendHttpRequest(verifiedRequest.method, verifiedRequest.url, verifiedRequest.payload)
+          .then((response) => {
```
Aschen, line 256: Nit: coding standards
```suggestion
          .then(response => {
```

src/protocols/WebSocket.ts
```diff
@@ -209,15 +211,61 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
...
+              .then(() => {
+                this.emit(verifiedRequest.payload.requestId, response);
+              });
+          })
+          .catch(error => this.emit(verifiedRequest.payload.requestId, {error}));
+      }
+      return;
     }
+
+    this.client.send(JSON.stringify(request));
```
Aschen, line 268: Can't this method return a rejected promise?

Aschen (changes requested):
Some comments, good job otherwise :+1:

scottinet (changes requested):

src/protocols/WebSocket.ts
Shiranuit: If there is something wrong this method will throw rejecting the call to query

src/protocols/WebSocket.ts
Shiranuit: No, if the offline queue is not enabled on Kuzzle constructor, then after the disconnection none of the requests will be queued.

Is there a case, in a well designed application, where someone would make a requests to `auth:login`, `auth:logout`, `auth:refreshToken` and then continue to make request without waiting for them to be resolved ?

Do you think that we should enable the offline queue when cookieAuth is enabled ?

doc/7/protocols/websocket/introduction/index.md
Shiranuit: I will just update the schema before adding it

src/protocols/WebSocket.ts
Aschen: The user does not expect his protocol to be disconnected and if he enabled the offline queue then he does not expect to loose any request.

IMHO during the cookie Auth process (ws disconnect, http request, ws connect), we should queue requests and replay them when the websocket connection is open again.

@scottinet Wdyt?

Shiranuit: I mean if he enabled offline queue in the kuzzle constructor everything is fine, nothing should be loose since after the disconnection the offline queue will queue the following requests and replay them after being reconnected, so this is fine for this part.

But I was wondering, isn't it an undefined behaviour to make requests after `auth:login` `auth:logout` or `auth:refreshToken` without waiting for the `auth` request to be resolved ?
example:
```js
kuzzle.auth.login(...); // login is not awaited
kuzzle.index.create('nyc-open-data'); // We cannot ensure that index.create is not sent before auth.login has been resolved
```
Am I right ?

Aschen: Yes, developers should await but I'm thinking about a request made in another part of the application since the login action will take few event loop turn.

For login I'm not sure if it is a problem  but for a refreshToken for example the developer expect to never lost his connection to Kuzzle since he is refreshing his token at fixed intervals.

The whole process involve deconnection of the websocket connection so if requests are not queued some data will be loss. IMHO we should activate the offline queue and play back the requests even if the SDK was not constructed with this option

doc/7/protocols/websocket/introduction/index.md
Shiranuit: NVM, we might want to do that later

Shiranuit:
Update:
When `cookieAuth` option is set to `true` in the Kuzzle constructor, this enables the autoResubscribe option to make sure that subscriptions are retrieved upon reconnection, along with `offlineMode=auto` to enable queuing and replaying the requests after the connection has been reestablished.

src/Kuzzle.ts
```diff
@@ -231,15 +231,12 @@ export class Kuzzle extends KuzzleEventEmitter {
     
     if (this._cookieAuthentication) {
       this.protocol.enableCookieSupport();
+      options.offlineMode = 'auto'; // Enable offline queue and replay
```
Aschen, line 234: Actually the offline mode should only be actived during the time the websocket is disconnected. (only if it was not already activated)

Aschen (changes requested):

src/Kuzzle.ts
Shiranuit: Done

What has changed:
During the process of renewing the connection, the websocket now emit two new events that are used internally
`websocketRenewalStart` & `websocketRenewalDone`, the first one is emitted just before the disconnection and the other just after websocket connection has been restablished.

Kuzzle adds a listener to those events when cookieAuth is set to `true`, when the event `websocketRenewalStart` is triggered, Kuzzle will save the configuration state of the `autoReplay`, `autoQueue` and `autoResubscribe` and set all three to  `true`,
once the event `websocketRenewalDone` is triggered, Kuzzle will restore the previous state of the  `autoReplay`, `autoQueue` and `autoResubscribe` configurations

Aschen (approved):

PR #622, opened 2021-03-26
Pick first route when multiple non-GET routes of the same size are available

scottinet:
# Description

When Kuzzle exposes multiple HTTP routes for the same controller action, the SDK picks either the one with the shortest URL and, if there is a tie in URL lengths, it prefers the GET route to favor content caching.

But when multiple routes are available, all with the same size, and none are a GET route, the SDK picks... none of them. Resulting in some actions returning a "URL not found" when trying to use them.

This PR fixes that.


MathieuVeber (approved):
Thanks :tada:

Aschen (approved):

PR #632, opened 2021-04-22
Add a timeout option for every request

Shiranuit:
## What does this PR do ?
This PR adds a default timeout to every request and a new option timeout to the method `query` and the actions of the following controllers:
- Auth (except for `login` & `logout`)
- Document
- Collection
- Index
- Realtime
- Bulk

This new option can be used to set a timeout (in ms) to every request sent, after the timeout if the  request has not been resolved yet the promise will be rejected with a message saying that the request has timed out.

It's also possible to configure a global requestTimeout in the kuzzle constructor.

If the timeout is set to `-1` the promise will never be reject and will be waited indefinitely.

### Boyscout

Reformatting of tables in markdown files


doc/7/controllers/auth/create-api-key/index.md
```diff
@@ -21,20 +21,21 @@ createApiKey(description, [options]);
...
-| `refresh` | <pre>boolean</pre><br />(`false`) | If set to `wait_for`, Kuzzle will not respond until the API key is indexed |
+| Property    | Type<br />(default)                  | Description                                                                                                           |
+| ----------- | ------------------------------------ | --------------------------------------------------------------------------------------------------------------------- |
+| `expiresIn` | <pre>string/number</pre><br />(`-1`) | Expiration duration                                                                                                   |
+| `_id`       | <pre>string</pre><br />(`null`)      | API key unique ID                                                                                                     |
+| `refresh`   | <pre>boolean</pre><br />(`false`)    | If set to `wait_for`, Kuzzle will not respond until the API key is indexed                                            |
+| `timeout`   | <pre>number</pre><br/>(`-1`)         | Time (in ms) during which a request will still be waited to be resolved. Set it `-1` if you want to wait indefinitely |
```
Leodau, line 38: `Set it -1 if you want to wait indefinitely`
Well, yeah but not really... -1 is the default right?

doc/7/controllers/auth/create-my-credentials/index.md
```diff
@@ -27,9 +27,10 @@ createMyCredentials(strategy, credentials, [options]);
 
 Additional query options
 
-| Property   | Type<br/>(default)              | Description                                                                  |
-| ---------- | ------------------------------- | ---------------------------------------------------------------------------- |
-| `queuable` | <pre>boolean</pre><br/>(`true`) | If true, queues the request during downtime, until connected to Kuzzle again |
+| Property   | Type<br/>(default)              | Description                                                                                                           |
+| ---------- | ------------------------------- | --------------------------------------------------------------------------------------------------------------------- |
+| `queuable` | <pre>boolean</pre><br/>(`true`) | If true, queues the request during downtime, until connected to Kuzzle again                                          |
+| `timeout`  | <pre>number</pre>               | Time (in ms) during which a request will still be waited to be resolved. Set it `-1` if you want to wait indefinitely |
```
Leodau, line 33: So at this action, this timeout has no default?

src/controllers/Auth.ts
```diff
@@ -121,8 +134,15 @@ export class AuthController extends BaseController {
...
    * @param options Additional options
    *    - `refresh` If set to `wait_for`, Kuzzle will not respond until the API key is indexed
+   *    - `timeout` Request Timeout in ms, after the delay if not resolved the promise will be rejected
    */
-  deleteApiKey(id: string, options: { refresh?: 'wait_for' } = {}): Promise<null> {
+  deleteApiKey(
+    id: string,
+    options: {
+      refresh?: 'wait_for',
+      timeout?: number
```
Leodau, line 143: In the future we should take a look at how to handle these 'shared' options for all queries, having 'refresh' and now 'timeout' could just be the beginning of having tons of query oriented options for every API action.

doc/7/controllers/auth/create-api-key/index.md
Shiranuit: There is a global default which can be configured with the `options` of the Kuzzle Constructor but is set to `-1` by default
Then each method can override this configuration on a per request basis.

doc/7/controllers/auth/create-my-credentials/index.md
Shiranuit: There is a default timeout but IDK how to explain that the default timeout  is  the global timeout configured in the Kuzzle constructor if it has been configured or `-1` otherwise, meaning that there is no timeout

src/controllers/Auth.ts
Shiranuit: You're right, I might be  a good thing to have a better way of handling this kind of shared params, this might need a Workshop

src/Kuzzle.ts
```diff
@@ -87,6 +87,7 @@ export class Kuzzle extends KuzzleEventEmitter {
   private _queueMaxSize: any;
   private _queueTTL: any;
   private _replayInterval: any;
+  private _requestTimeout: any;
```
Aschen, line 90: This should be typed propertly

src/Kuzzle.ts
```diff
@@ -632,7 +655,8 @@ export class Kuzzle extends KuzzleEventEmitter {
             resolve,
             reject,
             request,
-            ts: Date.now()
+            ts: Date.now(),
+            timeout: requestTimeout
```
Aschen, line 659: Missing last semicolon

src/Kuzzle.ts
```diff
@@ -815,4 +846,26 @@ Discarded request: ${JSON.stringify(request)}`));
 
     dequeuingProcess();
   }
+
+  /**
+   * Sends a request with a timeout
+   */
+  private _timeoutRequest(delay, {request, options = {}}) {
```
Aschen, line 853: This should be typed properly

src/Kuzzle.ts
```diff
@@ -815,4 +846,26 @@ Discarded request: ${JSON.stringify(request)}`));
...
+    // No timeout
+    if (delay === -1) {
+      return this.protocol.query(request, options);
+    }
+
+    const timeout = new Promise((resolve, reject) => {
+      const id = setTimeout(() => {
+        clearTimeout(id);
```
Aschen, line 861: I'm not sure you need to clear the timeout. Usually you do this to cancel a timeout before execution but once it's executed Node.js will clear the internal timers itself

src/Kuzzle.ts
```diff
@@ -815,4 +846,26 @@ Discarded request: ${JSON.stringify(request)}`));
...
+        reject(`Request timed out after ${delay} ms`);
```
Aschen, line 862: Rejection should be done with a proper error class. 
IMHO a new error class must be created

src/Kuzzle.ts
```diff
@@ -815,4 +846,26 @@ Discarded request: ${JSON.stringify(request)}`));
...
+      }, delay);
+    });
+
+    return Promise.race([
+      timeout,
+      this.protocol.query(request, options)
```
Aschen, line 868: Open debate / reflection:
Doing this will not cancel the original request, it may still be executed after a certain amount of time.
I don't think we can do something about it actually but at least we should add a warning in the documentation

test/kuzzle/query.test.js
```diff
@@ -6,6 +6,50 @@ const
...
+    });
+
+    it('should resolve the response when a response is obtained before the configured request timeout', async () => {
+      kuzzle.protocol.query = async () => { // Simulate that we are waiting the response to a query
+        return new Promise((resolve) => {
+          setTimeout(
+            () => {
+              resolve('foo');
+            },
+            1000
```
Aschen, line 29: Can we have smaller values to not slow down the whole test suit?
(apply for subsequent test cases)

Aschen (changes requested):

src/Kuzzle.ts
Shiranuit: That's right, the only thing this will do is prevent the SDK from being frozen if a request takes too long to be resolved, but Kuzzle will still resolve it at some point. If we really want to stop the request from being process after an amount of time, we should do it directly in Kuzzle and not in the SDK.

That said I don't know if that's what the issue was for

Any ideas ?

Aschen: Yes you are definitely right :+1: But even on Kuzzle side I don't think we can do much since we cannot remove a promise from the even loop.

I just thing we should add a warning in the documentation telling that requests are errored on the SDK side after the timeout but they still may be processed by Kuzzle

xbill82 (approved):

src/Kuzzle.ts
```diff
@@ -857,7 +857,7 @@ Discarded request: ${JSON.stringify(request)}`));
    * @param options Request options
    * @returns Resolved request or a TimedOutError
    */
-  private _timeoutRequest(delay: number, request:JSONObject, options:JSONObject = {}) {
+  private _timeoutRequest(delay: number, request:RequestPayload, options:JSONObject = {}) {
```
Aschen, line 860: Nitpicking: coding standard
```suggestion
  private _timeoutRequest(delay: number, request: RequestPayload, options: JSONObject = {}) {
```

Aschen (approved):

Leodau (approved):

PR #634, opened 2021-04-25
docs: use index from callback argument instead of indexOf

christophehurpeau:
## What does this PR do ?

Change in documentation: using `messages.indexOf(message)` is not necessary to get index because we can get it from callback 2nd argument


Aschen (approved):

Aschen:
Good catch @christophehurpeau !
Thanks for your contribution :-)

PR #635, opened 2021-04-28
Fix cookie authentication support for WebSocket

Shiranuit:
## What does this PR do ?

This PR change the behaviour of the websocket's `connect` method to call `HttpProtocol.connect()` if the support of cookieAuthentication is enabled.
Without that the HttpProtocol routes are not constructed, causing it to not send the payloads.

```bash
npm run test
```
### Other changes
Correction of a wrong statement in the cookieAuth option of the Kuzzle constructor.


Yoann-Abbes (approved):

alexandrebouthinon (approved):

scottinet (changes requested):
Missing package version bump (run `npm version patch` and commit the changes)

scottinet (approved):

PR #636, opened 2021-05-11
Add support for the document:deleteFields API action

Shiranuit:
## What does this PR do ?
This PR add the action `document:deleteFields` and the associated documentation and tests
⚠️ Depends on #632

### DIFF
--- a/doc/7/controllers/document/delete-fields/snippets/deleteFields.js
+++ b/doc/7/controllers/document/delete-fields/snippets/deleteFields.js
@@ -1,0 +1,7 @@
+try {
+  const response = await kuzzle.document.deleteFields('nyc-open-data', 'yellow-taxi', 'some-id', ['bar'], {source: true});
+
+  console.log(response._source);
+} catch (error) {
+  console.error(error.message);
+}
--- a/src/controllers/Document.ts
+++ b/src/controllers/Document.ts
@@ -188,6 +188,48 @@ export class DocumentController extends BaseController {
   }
 
   /**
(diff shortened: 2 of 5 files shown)


src/controllers/Document.ts
```diff
@@ -187,6 +187,48 @@ export class DocumentController extends BaseController {
...
+   *    - `silent` If true, then Kuzzle will not generate notifications
+   *    - `source` If true, the response will contain the updated document
+   *
+   * @returns The updated document
+   */
+  deleteFields(
+    index: string,
+    collection: string,
+    _id: string,
+    fields: Array<string>,
```
Aschen, line 210: Prefer the more concise `string[]`

Aschen (changes requested):
You should add the `timeout` option from your other PR

MathieuVeber (approved):
LGTM :wink:

Aschen (approved):

doc/7/controllers/document/delete-fields/snippets/deleteFields.js
```diff
@@ -0,0 +1,7 @@
+try {
+  const response = await kuzzle.document.deleteFields('nyc-open-data', 'yellow-taxi', 'some-id', ['bar'], {source: true});
```
Leodau, line 2: Nit
```suggestion
  const response = await kuzzle.document.deleteFields('nyc-open-data', 'yellow-taxi', 'some-id', ['bar'], { source: true });
```

Leodau (approved):

PR #645, opened 2021-06-07
Latest kuzdoc minor and dead-link check

xbill82:
Following the latest improvements in the Docs CI workflow, we need to have isomorphic dependencies (thus `npm ci`) but the latest Kuzdoc minor within the major specified by `package.json` (thus `npm i kuzdoc`). Having the latest minor of Kuzdoc enables us to easily deploy non-breaking changes to all the CIs without manually bump the Kuzdoc dependency (which would be the case if we just`npm ci`).

Also, this PR re-activates the dead-link check, using the SSH-key.

Sorry for the linting.

⚠️  Once merged, we should merge back to `7-dev`

### DIFF
--- a/.github/actions/dead-links/action.yml
+++ b/.github/actions/dead-links/action.yml
@@ -4,7 +4,7 @@ runs:
   using: "composite"
   steps:
     - name: Install deps
-      run: npm ci
+      run: npm ci --production=false
       shell: bash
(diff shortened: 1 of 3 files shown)


Yoann-Abbes (approved):

Leodau (approved):
Merged back to `7-dev`

Issue #647, opened 2021-07-06
core/Jwt expired getter return always true due to wrong comparaison (between timestamp in ms and sec)


robingrandval:
## Expected Behavior
Jwt expired getter should return true only if  'today' is after token expiration date

## Current Behavior
Always return true due to wrong comparaison. Date.now() return unix epoch time in ms and expiration time in jwt token payload is in seconde.

## Possible Solution
Convert Date.now() to secondes (cf. PR linked)

## Steps to Reproduce
1. call this getter after received a fresh JWT token, it return true (should be false)

## Context (Environment)

Kuzzle version: 2.11.2
Node.js version: 14.17.1
SDK version: 7.7.1

PR #648, opened 2021-07-06
Fix the Jwt.expired getter wrong comparison of micro timestamp and timestamp

robingrandval:
…ison (between timestamp in ms and sec)

Issue #647

### DIFF
--- a/src/core/Jwt.js
+++ b/src/core/Jwt.js
@@ -49,7 +49,7 @@ class Jwt {
   }
 
   get expired () {
-    return Date.now() > this.expiresAt;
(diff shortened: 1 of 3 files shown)


Aschen:
Hi @robingrandval and thanks for your contribution :-)

Can you add an unit test testing the `expired` getter when the JWT is not expired to ensure this bug will not reappear in the futur? https://github.com/kuzzleio/sdk-javascript/blob/master/test/core/Jwt.test.js#L42

robingrandval:
Hi @Aschen, it's done.

Aschen (approved):

alexandrebouthinon (approved):

PR #651, opened 2021-07-15
Hotfix heartbeat race condition

Shiranuit:
## What does this PR do ?

This PR fixes a Race Condition that was occuring within the Heartbeat causing the SDK to think the connection was lost and making him engage the reconnection protocol which was causing the SDK to create multiple Websocket connection even though the previous connections were fine.

This is a serious problem that could cause applications to be overloaded with Websocket connections and can saturate the Event Loop or the limit of Websocket connections defined by the browsers.

What has been done:
- Avoid creating a Timeout in the the ping Interval since they had the same amount of time before execution and that caused  a race condition, we now only have an Interval that checks if there has been a pong since the last cycle, if not throw an Error, if there was a pong since the last ping cycle do another ping.
- Correct tests to verify the new behaviour
- Fixed: the ping Interval was not cleared when `websocket.close` was called
- Ensures that the websocket connection is closed on timeout before engaging the reconnection protocol otherwise if this was a false positive because Kuzzle could not respond in time we would have 2 connection opened.
- Ensures that the previous interval is cleared before creating a new one

## How should this be manually tested
Run this script, there should not be any network error messages, and only one active websocket connection to kuzzle.
Unfortunately this doesn't help testing the `Race Condition` as it's not easy to reproduce the exact conditions needed, but it somehow occurs frequently in one of our project.
```js
const { Kuzzle, WebSocket } = require('./index');

const kuzzle = new Kuzzle(new WebSocket('localhost'));

async function main() {
    kuzzle.on('networkError', (...args) => {
        console.log(...args);
    });
    for (let i  = 0; i < 10; i++) {
        await kuzzle.disconnect();
        await kuzzle.connect();
    }
}

main();
```

### DIFF
--- a/src/protocols/WebSocket.ts
+++ b/src/protocols/WebSocket.ts
@@ -28,7 +28,7 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
    *    - `port` Kuzzle server port (default: `7512`)
    *    - `headers` Connection custom HTTP headers (Not supported by browsers)
    *    - `reconnectionDelay` Number of milliseconds between reconnection attempts (default: `1000`)
-   *    - `pingInterval` Number of milliseconds between two pings (default: `10000`)
+   *    - `pingInterval` Number of milliseconds between two pings (default: `2000`)
    *    - `ssl` Use SSL to connect to Kuzzle server. Default `false` unless port is 443 or 7443.
    */
   constructor(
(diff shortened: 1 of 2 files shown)


src/protocols/WebSocket.ts
```diff
@@ -108,23 +106,39 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
...
         */
+        clearInterval(this.pingIntervalId);
+        this.waitForPong = false; // Reset when connection is established
         this.pingIntervalId = setInterval(() => {
-          if (this.client && this.client.readyState === 1) {
+          // If the connection is established and we are not waiting for a pong we ping Kuzzle
+          if (this.client &&
+              this.client.readyState === this.client.OPEN &&
+              !this.waitForPong
+          ) {
```
Aschen, line 124: Wrong coding standards here, the `&&` should be at the beginning

Aschen (approved):

PR #653, opened 2021-07-27
Reconnect with authenticator when receiving TokenExpired event

Aschen:
## What does this PR do ?

When the SDK subscribe to realtime notification and the authentication expire, Kuzzle send a `TokenExpired` notification.  

When the SDK receive this notification, it unsubscribe from every realtime subscriptions.

A proper re-authentification mechanism should be implemented instead when the `authenticator` property has been set

### DIFF
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -591,11 +591,11 @@ export class Kuzzle extends KuzzleEventEmitter {
     if (this._reconnectInProgress) {
       return;
     }
-
+    
     if (this.autoQueue) {
       this.stopQueuing();
     }
-
+  
     // If an authenticator was set, check if a user was logged in and  if the token is still valid and try
     // to re-authenticate if needed. Otherwise the SDK is in disconnected state.
     if ( this._loggedIn
@@ -838,6 +838,7 @@ Discarded request: ${JSON.stringify(request)}`));
       return;
     }
 
+    console.log(this._loggedIn);
     if (this._loggedIn && this.authenticator && await this.tryReAuthenticate()) {
       this.emit('reAuthenticated');
 
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -838,7 +838,6 @@ Discarded request: ${JSON.stringify(request)}`));
       return;
     }
 
-    console.log(this._loggedIn);
     if (this._loggedIn && this.authenticator && await this.tryReAuthenticate()) {
       this.emit('reAuthenticated');
 
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -589,8 +589,6 @@ export class Kuzzle extends KuzzleEventEmitter {
       return;
     }
 
-    this._reconnectInProgress = true;
-
     if (this.autoQueue) {
       this.stopQueuing();
     }
@@ -601,7 +599,6 @@ export class Kuzzle extends KuzzleEventEmitter {
       && this.authenticator 
       && ! await this.tryReAuthenticate()
     ) {
-      this._reconnectInProgress = false;
       this._loggedIn = false;
       this.disconnect();
       
@@ -613,8 +610,6 @@ export class Kuzzle extends KuzzleEventEmitter {
     }
     
     this.emit('reconnected');
-
-    this._reconnectInProgress = false;
   }
 
   /**
@@ -628,6 +623,7 @@ export class Kuzzle extends KuzzleEventEmitter {
    * This method never returns a rejected promise.
    */
   private async tryReAuthenticate (): Promise<boolean> {
+    this._reconnectInProgress = true;
     try {
       const { valid } = await this.auth.checkToken();
 
@@ -645,6 +641,8 @@ export class Kuzzle extends KuzzleEventEmitter {
       });
 
       return false;
+    } finally {
+      this._reconnectInProgress = false;
     }
   }
 
@@ -833,6 +831,9 @@ Discarded request: ${JSON.stringify(request)}`));
    * Throttles to avoid duplicate event triggers.
    */
   async tokenExpired () {
+    if (this._reconnectInProgress) {
+      return;
+    }
 
     if (this._loggedIn && this.authenticator && await this.tryReAuthenticate()) {
       this.emit('reAuthenticated');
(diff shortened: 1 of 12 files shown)


src/Kuzzle.ts
```diff
@@ -769,7 +770,13 @@ Discarded request: ${JSON.stringify(request)}`));
    * On token expiration, reset jwt and unsubscribe all rooms.
    * Throttles to avoid duplicate event triggers.
    */
-  tokenExpired() {
+  async tokenExpired () {
+    if (this.authenticator && await this.tryReAuthenticate()) {
```
Aschen, line 774: We still have an issue because if the user call `auth.logout`, the `tokenExpired` event will be triggered and thus the authenticator will be used even if the user didn't wanted to re-auth

Shiranuit: We might need to have a state when we successfully login and set it to false when doing logout

Shiranuit:
## What has been done

- [x] Fix `loginAttempt` event not being emitted when login has failed in cookieAuth mode
- [x] Added a new event called `logoutAttempt` that is emitted on logout the same way as `loginAttempt`
- [x] Kuzzle now store a state `_loggedIn` that is usefull to know if the SDK was previously logged in or not
- [x] Avoid reconnection loop when the SDK is in cookieAuth mode that were triggered by the socket renewal procedure
- [x] ReAuthentication using `authenticator` is now only done when we were previously logged in

## Goal

The ultimate goal of those changes is to only do a reauthentication when we were previously logged in, this is possible thanks to the events `loginAttempt` and `logoutAttempt` that can be used to know when a logout or login have been successfuly accomplished.
Thanks to those new event Kuzzle can then store a state that helps know if we were logged in.

### Why doing a `checkToken` when the  `connected` event is emitted ?

Simply because if we were using cookies to store our Token, the SDK will use it directly in every request and if the Token is still valid, every requests will be executed with the right of the user that the Token belong to, without the needs to do a `login` before.
This why we should check the validity of the token at the connection, because if the Token is still valid it's the same as doing a `login` before sending requests.

### Why doing a `checkToken` if the `loginAttempt` has failed instead of just setting the `_loggedIn` value to `false` ?

This is also because of the cookieAuth mode, if a previous token is stored in the browser's cookies and we then want to login to another account and the login fails, the SDK will still use the previous token and we should verify the token validity.

Shiranuit:
## Other Enhancements

### Reconnection issue

There was a bug were the reconnection process was executed twice because calling `websocketClient.close()` in the HeartBeat to close the previous Websocket do call the `websocketClient.onclose` callback that was calling  `clientNetworkError` since the closing status code was not `1000` this combined with the fact that the HeartBeat was also calling `clientNetworkError`, the method `clientNetworkError` was called twice.

`clientNetworkError` is the method called on connection lost, responsible of starting the reconnection mecanism.

This was the cause of a double `reconnected` event emitted.

Now the `reconnected` event is only emitted once.

### Disconnected event emitted multiples times

I noticed that the `disconnected` was emitted multiple times while the SDK was trying to reconnect after a connection loss.
This behaviour was caused by the `clientNetworkError` emitting a `disconnected` event each time the connection retry had fail instead of only emitting the event once when a previously established connection was lost.

Now the `disconnected` event is only emitted once, when a previously established connection is lost.
The `networkError` event is still emitted at each reconnection attempt because it make sense to know what is the reason that has prevented the establishment of a new connection.

Njuelle (approved):

Yoann-Abbes (approved):

src/controllers/Realtime.ts
```diff
@@ -60,6 +60,10 @@ export class RealtimeController extends BaseController {
     this.kuzzle.on('disconnected', () => this.saveSubscriptions());
     this.kuzzle.on('networkError', () => this.saveSubscriptions());
     this.kuzzle.on('reconnected', () => this.resubscribe());
+    this.kuzzle.on('reAuthenticated', () => {
+      this.saveSubscriptions();
```
alexandrebouthinon, line 64: Why do you save the subscriptions? I understand you want to keep them safe in case of error during reAuth but I think this could lead to subscriptions duplication if reAuth failed then success. Don't you think so?

src/protocols/WebSocket.ts
```diff
@@ -292,14 +273,16 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
             this.emit('websocketRenewalDone'); // Notify that the websocket has finished renewing his connection with Kuzzle
           });
       })
-      .catch(error => this.emit(formattedRequest.payload.requestId, {error}));
-
+      .catch(error => {
```
alexandrebouthinon, line 276: nit: should we asyncify this function call?

test/kuzzle/authenticator.test.js
```diff
@@ -49,11 +147,32 @@ describe('Kuzzle authenticator function mecanisms', () => {
...
+      promise = new Promise(_resolve => {
+        resolve = _resolve;
+      });
+      await kuzzle.connect();
+
+      kuzzle._loggedIn = false;
+
+      protocol.emit('reconnect');
+
+      // We need a timeout since the listener on "reconnect" even is async
```
alexandrebouthinon, line 160:
```suggestion
      // We need a timeout since the listener on "reconnect" event is asynchronous
```

test/kuzzle/authenticator.test.js
```diff
@@ -31,12 +127,14 @@ describe('Kuzzle authenticator function mecanisms', () => {
...
       promise = new Promise(_resolve => {
         resolve = _resolve;
       });
       await kuzzle.connect();
 
+      kuzzle._loggedIn = true;
+
       protocol.emit('reconnect');
 
       // We need a timeout since the listener on "reconnect" even is async
```
alexandrebouthinon, line 140:
```suggestion
      // We need a timeout since the listener on "reconnect" event is asynchronous
```

alexandrebouthinon (changes requested):

src/controllers/Realtime.ts
Shiranuit: This will not lead to subscription duplication since this event is called after the authenticator is called, meaning that an `auth:login` operation has been done.
Since a `auth:login` has been done the subscriptions are not existing anymore in Kuzzle, we do the save the subscriptions of the SDK to then do a resubscribe directly to renew the subscriptions.

src/protocols/WebSocket.ts
Shiranuit: I dont see why

alexandrebouthinon (approved):

Shiranuit (approved):

PR #657, opened 2021-08-09
Automatic resubscription when logged in using jwt property

Shiranuit:
## What does this PR do ?

This PR enables the automatic resubscription of the SDK when logged in using the JWT property

### DIFF
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -430,12 +430,7 @@ export class Kuzzle extends KuzzleEventEmitter {
   set jwt (encodedJwt) {
     this.auth.authenticationToken = encodedJwt;
 
-    if (! encodedJwt) {
-      this._loggedIn = false;
-      return;
-    }
-
-    this._loggedIn = true;
+    this._loggedIn = encodedJwt ? true : false;
   }
 
   get offlineQueue () {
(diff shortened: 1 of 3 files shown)


src/Kuzzle.ts
```diff
@@ -429,6 +429,13 @@ export class Kuzzle extends KuzzleEventEmitter {
 
   set jwt (encodedJwt) {
     this.auth.authenticationToken = encodedJwt;
+
+    if (! encodedJwt) {
```
Aschen, line 433: Nitpicking: a ternary could be easier to read here

Aschen (approved):

src/Kuzzle.ts
```diff
@@ -429,6 +429,8 @@ export class Kuzzle extends KuzzleEventEmitter {
 
   set jwt (encodedJwt) {
     this.auth.authenticationToken = encodedJwt;
+
+    this._loggedIn = encodedJwt ? true : false;
```
xbill82, line 433: shouldn't this be a getter checking the existence of `this.auth.authenticationToken` instead of a static member synced by another setter?

xbill82 (approved):

src/Kuzzle.ts
Shiranuit: I can't because in cookieAuth mode there is no JWT property set but that doesn't mean the SDK is not logged out, that why I cannot always check if there is a JWT but I can set a property when people do want to put a JWT by hand.

alexandrebouthinon (approved):

PR #658, opened 2021-08-10
Resubscription not working in the browser because of a bad transpilation

Shiranuit:
## What does this PR do ?

This PR make a slight changes in the way we loop over the `Map.keys()` in the `saveSubscription`, `resubscribe` and `removeSubscriptions` method, because Webpack is incorrectly transpiling the loops making those methods not working properly.

Instead of iterating this way
```js
for (const roomId of this._subscriptions.keys()) {
  // Body
}
```
The iteration is now done using `Map.forEeach`
```js
this._subscriptions.forEach((rooms, roomId) => {
  // Body
});
```

### Why those changes

Like I said, those changes are made to fix an issue where the resubscription mecanism wasn't working in the JS SDK but only in the browser.
This was caused by a bad transpilation of the loops using Webpack:
```js
for (const roomId of this._subscriptions.keys())
```
That were transpiled to:
```js
for(let e=0,t=this._subscriptions.keys();e<t.length;e++)
```

Since `_subscriptions` is a `Map` object, and `Map.keys()` return a `MapIterator` that doesn't have the property `length`, the loop wasn't executed since the condition was invalid.

### DIFF
--- a/src/controllers/Realtime.ts
+++ b/src/controllers/Realtime.ts
@@ -279,7 +279,7 @@ export class RealtimeController extends BaseController {
      * Use forEach instead of iterating over Map.keys() because the Webpack 
      * transpilation is producing bad code leading to a loop not iterating.
      */
-    this._subscriptionsOff.forEach((rooms, roomId) => {
+    this._subscriptionsOff.forEach((rooms) => {
       for (const room of rooms) {
         room.removeListeners();
       }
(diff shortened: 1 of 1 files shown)


alexandrebouthinon (approved):

Aschen (approved):

xbill82 (approved):

xbill82:
Looks like it's not a bug from Webpack but rather Babel (or any other transpiler you are using through Webpack). Are we considering submitting this issue to the author of the transpiler?

Shiranuit:
> Looks like it's not a bug from Webpack but rather Babel (or any other transpiler you are using through Webpack). Are we considering submitting this issue to the author of the transpiler?

You're right I made a shortcut here, we're using babel, for now I didn't tried to research whether that was a known bug or not, but it might be good to make an issue if that was not already addressed.

PR #663, opened 2021-09-16
Added security types

inform880:
## What does this PR do ?

This PR converts the security.js to a TypeScript file in order to get autocompletion.

  - Step 1 :
  - Step 2 :
  - Step 3 :


Aschen (changes requested):
Thanks for your contribution!

Can you remove the original `SecurityController.js` file from the repo since it's a generated file now please?

Also, you should add this file to the [.eslintignore](https://github.com/kuzzleio/sdk-javascript/blob/master/.eslintignore) and [.gitignore](https://github.com/kuzzleio/sdk-javascript/blob/master/.gitignore) files.

inform880:
@Aschen Do I have to worry about some of the checks failing, not sure if I have more changes I have to make?

Aschen:
> @Aschen Do I have to worry about some of the checks failing, not sure if I have more changes I have to make?

No it's because your PR came from a fork so the of the secrets we are using in our CI are not present.

PR #664, opened 2021-09-20
Fix reconnectionError not emittend when authenticator is missing

Shiranuit:
## What does this PR do ?

Emit a reconnectionError when there is no authenticator set.

### DIFF
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -640,6 +640,9 @@ export class Kuzzle extends KuzzleEventEmitter {
        * like so API Keys can be used even if there is no authenticator since they will be still valid.
        */
       if (! this.authenticator) {
+        this.emit('reconnectionError', {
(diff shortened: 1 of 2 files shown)


src/Kuzzle.ts
```diff
@@ -640,6 +640,9 @@ export class Kuzzle extends KuzzleEventEmitter {
        * like so API Keys can be used even if there is no authenticator since they will be still valid.
        */
       if (! this.authenticator) {
+        this.emit('reconnectionError', {
+          error: new Error('Could not reauthenticate: undefined authenticator property')
```
Aschen, line 644: Wording nitpicking
```suggestion
          error: new Error('Could not re-authenticate: "authenticator" property is not set.')
```

Aschen (approved):

Aschen (approved):

PR #668, opened 2021-10-03
Propagates every arguments to Kuzzle

Aschen:
## What does this PR do ?

This PR allows the SDK to propagate every arguments passed to a method to Kuzzle even if this one is not explicitly declared.

Example
```js
await sdk.document.get('index', 'collection', 'document-id', { myCustomArg: 'foobar' });

// Generated query for Kuzzle API
{
  controller: 'document', 
  action: 'get',
  index: 'index',
  collection: 'collection',
  _id: 'document-id',
  myCustomArg: 'foobar',
}
```

### How the code has been modified

Since I'm lazy and I didn't wanted to modify every methods one by one to extract an interface from the inline type definition, I used the [ts-morph](https://ts-morph.com/) package who allows to directly manipulate Typescript AST.

You can check the `ast-refactor/inline-options-to-interface.ts` if you are curious about how it has been made.

### DIFF
--- a/ast-refactor/inline-options-to-interface.ts
+++ b/ast-refactor/inline-options-to-interface.ts
@@ -1,0 +1,111 @@
+/**
+ * Script to refactor inline options declaration into external interfaces
+ * inheriting from the default "ArgsDefault" interface.
+ * 
+ * @example
+ * 
+export class SomeController {
+  helloWorld (
+    name: string,
+    options: {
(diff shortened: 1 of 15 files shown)


xbill82 (approved):
🤖 I LIEK TIHS COMPUT3R PROGRAM VERY MUCH

rolljee (approved):

PR #695, opened 2022-04-13
Fix Bad links in the Kuzzle event documentation

clementbolin:
## What does this PR do ?

Fix bad links in the documentation


Aschen (approved):

Aschen:
@clementbolin PR fixing the documentation can be done directly on master so they will deploy the production version of the documentation when merged

PR #696, opened 2022-04-13
Add `props` in KuzzleError

Aschen:
## What does this PR do ?

Kuzzle returns a `props` property with error, this property contains all the string used to build the error message.

Those string can be used by the application layer to build other error messages.

This PR expose this property

### DIFF
--- a/src/protocols/abstract/Base.ts
+++ b/src/protocols/abstract/Base.ts
@@ -152,7 +152,6 @@ Discarded request: ${JSON.stringify(request)}`));
       this._pendingRequests.delete(request.requestId);
 
       if (response.error) {
-        console.log(response.error)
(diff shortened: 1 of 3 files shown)


src/protocols/abstract/Base.ts
```diff
@@ -152,6 +152,7 @@ Discarded request: ${JSON.stringify(request)}`));
       this._pendingRequests.delete(request.requestId);
 
       if (response.error) {
+        console.log(response.error)
```
Shiranuit, line 155:
```suggestion
```
Debug leftover ?

clementbolin: not a `console.error()` ?

clementbolin (approved):

Issue #682, opened 2022-01-20
Ensure that the callback passed to realtime can not cause unhandledRejection


Aschen:
## Feature Description

The callback function passed to the `realtime.subscribe` method can returns a promise and if this promise is rejected then we end up with an unhandled rejection.

The callback could be awaited and any error catch, a corresponding event should be emitted with the error

https://github.com/kuzzleio/sdk-javascript/blob/0a1e26b4f5961061804567bc67bc9fd814b3ffa6/src/core/Room.js#L83

PR #699, opened 2022-04-20
Add event at callback error in subscription

ChillPC:
#682 

## What does this PR do ?

Add event at callback error in subscription.

An new event type has been created.

## How to test

Considering `index1:collection1` exist, this code:

```js
const {
  Kuzzle,
  WebSocket,
} = require('kuzzle-sdk');

const kuzzle = new Kuzzle(
  new WebSocket('localhost')
);

kuzzle.on('networkError', error => {
  console.error('Network Error:', error);
});

const run = async () => {
  try {
    await kuzzle.connect();

    console.log('--- Subscribe to event');
    kuzzle.on(
      'callbackError',
      (error) => console.log(
        `Error message "${error.error.message}" on "${error.room.index}:${error.room.collection}"`)
    );

    const cb = () => {throw new Error("I am a callback and I threw an error")};
    const fieldA = 'callbackErrorTest';
    const filter = {
      equals: {
        fieldA
      }
    };

    console.log('--- Subscribe with callback');
    await kuzzle.realtime.subscribe('index1', 'collection1', filter, cb);

    const newDocument = { fieldA };
    await kuzzle.document.create('index1', 'collection1', newDocument);

  } finally {
    kuzzle.disconnect();
  }
};

run();
```

Should log:

```js
--- Subscribe to event
--- Subscribe with callback
Error message "I am a callback and I threw an error" on "index1:collection1"
```

### DIFF
--- a/doc/7/essentials/events/index.md
+++ b/doc/7/essentials/events/index.md
@@ -118,13 +118,13 @@ Triggered whenever Kuzzle responds with an error
 
 ## callbackError
 
-Triggered whenever a callback on an event throw an error
+Triggered whenever the notification handler's callback returns a rejected promise
 
 **Callback arguments:**
 
 `@param {Error} error - Error details`
 
-`@param {Room} room - Room containing the callback`
+`@param {Notification} notification - Notification of the handler`
 
 ## reconnected
 
--- a/doc/7/essentials/events/index.md
+++ b/doc/7/essentials/events/index.md
@@ -116,6 +116,16 @@ Triggered whenever Kuzzle responds with an error
 
 `@param {object} request - Request that caused the error`
 
+## callbackError
+
+Triggered whenever a callback on an event throw an error
+
+**Callback arguments:**
+
+`@param {Error} error - Error details`
+
+`@param {Room} room - Room containing the callback`
+
 ## reconnected
 
 Triggered when the current session has reconnected to Kuzzle after a disconnection, and only if `autoReconnect` is set to `true`.
--- a/src/Kuzzle.ts
+++ b/src/Kuzzle.ts
@@ -73,6 +73,7 @@ export class Kuzzle extends KuzzleEventEmitter {
    * List of every events emitted by the SDK.
    */
   public events = [
+    'callbackError',
     'connected',
     'discarded',
     'disconnected',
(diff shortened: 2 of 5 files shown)


doc/7/essentials/events/index.md
```diff
@@ -116,6 +116,16 @@ Triggered whenever Kuzzle responds with an error
 
 `@param {object} request - Request that caused the error`
 
+## callbackError
+
+Triggered whenever a callback on an event throw an error
```
Aschen, line 121: This event should be emitted only when the notification handler returns a rejected promise

src/core/Room.js
```diff
@@ -80,7 +80,12 @@ class Room {
       data.volatile && data.volatile.sdkInstanceId === this.kuzzle.protocol.id;
 
     if (this.subscribeToSelf || !fromSelf) {
-      this.callback(data);
+      try {
+        this.callback(data);
```
Aschen, line 84: A rejected promise returned by the callback will still trigger an unhandledRejection here

Aschen (changes requested):
It's actually the user responsibility to catch and handle errors in the callback provided to `realtime.subscribe`.  

However, we want to maintain application stability and thus we should try to prevent unhandledRejection (when a promise is rejected and there is no catch handler nor await with try..catch).

The purpose of this issue is to handle any rejected promise that may be returned by the notification handler (the callback) and trigger an event when it's happen.

Aschen (changes requested):

test/core/room.test.js
```diff
@@ -333,6 +323,20 @@ describe('Room', () => {
...
+
+      cb.rejects(callbackError);
+
+      await room._channelListener(data);
+
+      should(cb)
+        .be.calledOnce()
+        .be.calledWith(data);
+
+      should(controller.kuzzle.emit).be.called();
```
Aschen, line 339: It could be nice check that the event contain the appropriate payload

src/Kuzzle.ts
```diff
@@ -86,6 +86,7 @@ export class Kuzzle extends KuzzleEventEmitter {
     'reconnected',
     'reconnectionError',
     'tokenExpired',
+    'callbackError',
```
Aschen, line 89: Nit: Can you sort this please?

Aschen (changes requested):

Aschen (approved):

PR #701, opened 2022-07-05
Use PATCH for `document:update` API action using HTTP protocol

alexandrebouthinon:
## What does this PR do ?
According to [this message](https://github.com/kuzzleio/kuzzle/blob/master/lib/api/httpRoutes.js#L370) and to avoid error in production context. We sould use PATCH HTTP verb instead of PUT for the `document:update` API action


Aschen (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-sdk",
-  "version": "7.10.0",
+  "version": "7.10.1",
```
TisserantG, line 3: The PR contain a very important breaking change, and will cause very important problems on all projects that use kuzzle with http if they upgrade their version without reading release note. 
And that is bad with a PATCH version. 
The only way to follow the semvers standard is to increment the major version or to increment minor and authorize PUT and PATCH (to avoir breaking change)

TisserantG (approved):

Issue #706, opened 2022-08-23
Add strong typing on the `query` method


Aschen:
## Feature Description

In order to comply with interfaces contracts defined as types, we need to add optional strong typing to the `query` method.

Each controller action should define it's own contract with 2 interfaces:
 - `ControllerActionRequest`
 - `ControllerActionResult`

Then those types can be used with the `query` method to ensure contracts are validated.

```
export interface NotificationSmsRequest extends RequestPayload {
  controller: 'notification',
  action: 'sms',
  engineId: string;
  body: {
    smsContent: string;
  }
};

interface NotificationSmsResult {
    smsCount: number;
}

const { result } = await sdk.query<NotificationSmsRequest, NotificationSmsResult>({
  controller: 'notification',
  action: 'sms',
  engineId: 'tenant-worksite-kuzzle',
  body: {
    smsContent: 'Hello, world',
  }
});

result.smsCount; // number
```

OlivierCavadenti:
Solve by https://github.com/kuzzleio/sdk-javascript/pull/707

PR #707, opened 2022-08-23
Add query typing

Aschen:
## What does this PR do ?

Implements https://github.com/kuzzleio/sdk-javascript/issues/706

A new `BaseRequest` interface can be extended to define the request type.

```js
export interface NotificationSmsRequest extends BaseRequest {
  controller: 'notification',
  action: 'sms',
  engineId: string;
  body: {
    smsContent: string;
  }
};

interface NotificationSmsResult {
    smsCount: number;
}

const { result } = await sdk.query<NotificationSmsRequest, NotificationSmsResult>({
  controller: 'notification',
  action: 'sms',
  engineId: 'tenant-worksite-kuzzle',
  body: {
    smsContent: 'Hello, world',
  }
});

result.smsCount; // number
```

### DIFF
--- a/src/types/RequestPayload.ts
+++ b/src/types/RequestPayload.ts
@@ -5,16 +5,16 @@ import { JSONObject } from './JSONObject';
  *
  * @see https://docs.kuzzle.io/core/2/api/payloads/request
  */
-export type RequestPayload = {
+export interface RequestPayload {
   /**
    * API controller name
    */
-  controller?: string;
+  controller: string;
 
   /**
    * API action name
    */
-  action?: string;
+  action: string;
 
   /**
    * Index name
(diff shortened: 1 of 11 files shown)


src/types/RequestPayload.ts
```diff
@@ -5,16 +5,16 @@ import { JSONObject } from './JSONObject';
  *
  * @see https://docs.kuzzle.io/core/2/api/payloads/request
  */
-export type RequestPayload = {
+export interface RequestPayload {
```
OlivierCavadenti, line 8: Don't forget to update this page: https://docs.kuzzle.io/core/2/framework/types/request-payload/

Aschen: Unfortunately, this is in Kuzzle doc :roll_eyes: I will open a separate PR

OlivierCavadenti (approved):

PR #711, opened 2022-10-13
Fix bug where calling next on search with a from always return the same results

Shiranuit:
## What does this PR do ?

Fix bug where calling next on search with a from always return the same results

Prior to this change the SDK was always returning the same result each time the `next()` was called because the search was overriding the incremented `from` property with the given `from` property in the options causing each call to `next` to do a search from the begining
```js
const KuzzleSDK = require('kuzzle-sdk');

const kuzzle = new KuzzleSDK.Kuzzle(
  new KuzzleSDK.WebSocket(kuzzleHost),
);

(async () => {
  await kuzzle.connect();

  let results = await kuzzle.document.search('test','test',{},{from: 1, size:1}); 
    
  while (results) {
    console.log('res', results.hits.map(r=>r._id))
    results = await results.next();
  }
})();
```

### DIFF
--- a/src/protocols/abstract/Realtime.ts
+++ b/src/protocols/abstract/Realtime.ts
@@ -83,11 +83,12 @@ export abstract class BaseProtocolRealtime extends KuzzleAbstractProtocol {
       this.retrying = true;
 
       if (
+        window !== null &&
         typeof window === "object" &&
-        typeof window.navigator === "object" &&
-        window.navigator.onLine === false
+        typeof window!.navigator === "object" &&
+        window!.navigator.onLine === false
       ) {
-        window.addEventListener(
+        window!.addEventListener(
           "online",
           () => {
             this.retrying = false;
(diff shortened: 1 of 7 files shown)


alexandrebouthinon (approved):

src/protocols/abstract/Realtime.ts
```diff
@@ -83,11 +83,12 @@ export abstract class BaseProtocolRealtime extends KuzzleAbstractProtocol {
       this.retrying = true;
 
       if (
+        window !== null &&
```
Aschen, line 86: I don't really understand where is the actual fix. It seems that this line is the only addition to the codebase :thinking:

Shiranuit: Fixing the issue did not require an addition but a deletion.
The `from` and `size` parameters were sent both in the body and in the query params via the `options` property of query,  the catch is that the values in the body were different that the ones in the query params, the values in the query params are directly copied from the original query (because we wanted to keep the same options as the original query) but this had the effect to override the new values that were put in the body, causing the request to always be the same no matter what has been put in the body.

PR #716, opened 2022-12-21
Allow to call custom GET HTTP method with index and collection args

### LINKED ISSUE #713: Cannot access request argument with HTTP protocol on GET route
## Description

`index`, `collection` and other arguments are not present in request input while using the HTTP protocol of the SDK.

I suspect the URL to be incorrectly created by the SDK Http protocol since the generated route is `GET`

### Backend

```js

app.controller.register("efficient", {
  actions: {
    "efficient": {
      handler: async (request: KuzzleRequest) => {
console.log(request.input.args);
if (request.input.args.index === undefined) {
  throw new Error('No index');
}
      }
    }
  }
})
```

## Client (SDK)

```js
import { Kuzzle, Http } from 'kuzzle-sdk';

const kuzzle = new Kuzzle(new Http('localhost'));

await kuzzle.query({
    controller: 'efficient',
    action: 'efficient',
    index: "index",
    collection: "collection",
    body: { hey: "hey" },
});
```

alexandrebouthinon:
# Why?
Intend to fix #713 by adding a edge case handling in the request [payload forging loop](https://github.com/kuzzleio/sdk-javascript/blob/6aaf6e1630187a885191fd57123b7b25d946fb40/src/protocols/Http.ts#L264-L283) that allow user to call a custom HTTP GET route that expects to receive queries named index and/or collection.

### DIFF
--- a/src/protocols/Http.ts
+++ b/src/protocols/Http.ts
@@ -277,7 +277,10 @@ export default class HttpProtocol extends KuzzleAbstractProtocol {
       } else if (key === "index" || key === "collection") {
         // If we're calling a non-native route that answer to a GET request
         // we need to add the index and collection (if provided) to the query string
-        if (!staticHttpRoutes[request.controller][request.action] && method === "GET") {
+        if (
+          !staticHttpRoutes[request.controller][request.action] &&
+          method === "GET"
+        ) {
           queryArgs[key] = value;
         } else {
           payload[key] = value;
--- a/src/protocols/Http.ts
+++ b/src/protocols/Http.ts
@@ -274,6 +274,14 @@ export default class HttpProtocol extends KuzzleAbstractProtocol {
         payload.headers.authorization = "Bearer " + value;
       } else if (key === "volatile") {
         payload.headers["x-kuzzle-volatile"] = JSON.stringify(value);
+      } else if (key === "index" || key === "collection") {
+        // If we're calling a non-native route that answer to a GET request
+        // we need to add the index and collection (if provided) to the query string
+        if (!staticHttpRoutes[request.controller][request.action] && method === "GET") {
+          queryArgs[key] = value;
+        } else {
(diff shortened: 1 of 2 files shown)


Aschen (approved):
LGTML :tada:

src/protocols/Http.ts
```diff
@@ -274,6 +274,14 @@ export default class HttpProtocol extends KuzzleAbstractProtocol {
         payload.headers.authorization = "Bearer " + value;
       } else if (key === "volatile") {
         payload.headers["x-kuzzle-volatile"] = JSON.stringify(value);
+      } else if (key === "index" || key === "collection") {
```
Aschen, line 277: According to your comment [here](https://github.com/kuzzleio/sdk-javascript/issues/713#issuecomment-1362222584), shouldn't you handle other arguments like `meta` or `headers` ?

Aschen (commented):
According to your comment [here](https://github.com/kuzzleio/sdk-javascript/issues/713#issuecomment-1362222584), shouldn't you handle other arguments like `meta` or `headers` ?

alexandrebouthinon:
> According to your comment [here](https://github.com/kuzzleio/sdk-javascript/issues/713#issuecomment-1362222584), shouldn't you handle other arguments like `meta` or `headers` ?

Yes, it was on my mind. I think the `meta` support is necessary but for `headers` I'm not sure. I think there are too many internal mechanics that manipulate this field and I'm afraid to include a regression without wanting it. I think, if you agree, to support the `meta` field and put a deprecated on top of the request boilerplate section used to forge the final HTTP request. I think this system adds a lot of complexity and should be replaced in the next major version. What do you think?

Shiranuit (approved):

Aschen:
@alexandrebouthinon Let's fix only this for now, I agree :+1:

PR #723, opened 2023-02-21
Improve typings

ScreamZ:
Hi[^contact],
[^contact]: You can contact me on Discord if you have any questions and expect a quick answer. I'm already listed as a Contributor on it 😅.

Working on a new project with Kuzzle, I'm fixing some code to improve user experience.

**ℹ️ Note:** I'll keep updating this PR whenever I add new types (probably some in the following weeks).

# ⏩ Initial changes
## Notification type improvement

Improved the Notification type for real-time notification in order to work with type inference.

<details>
  <summary>Read details</summary>

### Before
```ts
kuzzle.realtime.subscribe("index", "messages", {}, (notification) => {
  // Only notification.type available
  if (notification.type === "document") {
    // notification still has no more properties…
  }
});
```
Before `notification` only contains the `type` property which is good because at this step we don't know the kind of notification. To continue further we need to type guard-safe this property in order to get further autocompletion.

To assert that we can make a custom guard or we can just let typescript deduct coherence with something like `notification.type === "document"`

In the previous version due to typing, this was not possible.

### After

Now TS is able to check the type correctly and infer that the notification is a `DocumentNotification` type.

```ts
kuzzle.realtime.subscribe("index", "messages", {}, (notification) => {
  // Only notification.type available
  if (notification.type === "document") {
    // notification is now inferred as a `DocumentNotification` and as access to every props of this type.
  }
});
```

</details>

## JSONObject Type
Typescript has a native type to deal with JSONObject `Record`. Also I could only recommend to check at [this](https://github.com/sindresorhus/type-fest/blob/main/source/basic.d.ts) and more generally about `type-fest` where Sindresorhus the author is a well-known and serious contributor over the internet. Why Not included `type-fest` directly as a deps ?

## Events for KuzzleEventEmitter
This adds a proper overload for KuzzleEventEmitter.
@Aschen I'm just not sure about `RequestPayload` type, because the doc at [this URL](https://docs.kuzzle.io/sdk/js/7/essentials/events) was not really explicit.

# ⏩ 28/02/23 Update

## Added coherence type checking test
`test/kuzzle-sdk-test.ts` is a file waiting for the test battery to be rewritten maybe someday in TypeScript.[^DefinitelyTyped]
[^DefinitelyTyped]: This is based on the same concept as described [here](https://github.com/DefinitelyTyped/DefinitelyTyped#my-package-teststs)

## First pass on typing `KuzzleEventEmitter.emit` and other internal first functions
I've been forced to update accordingly `emit` function with `unknown` in order to let `tsc` pass. Someday could be nice to add strong typing on emit too in order to avoid breaking the `on` API.

### DIFF
--- a/src/core/KuzzleEventEmitter.ts
+++ b/src/core/KuzzleEventEmitter.ts
@@ -1,8 +1,10 @@
+import { DisconnectionOrigin } from "../protocols/DisconnectionOrigin";
 import { KuzzleError } from "../KuzzleError";
 import { Notification, RequestPayload } from "../types";
 
+type ListenerFunction = (...args: unknown[]) => unknown;
 class Listener {
-  public fn: (...any) => any;
+  public fn: ListenerFunction;
   public once: boolean;
 
   constructor(fn, once = false) {
@@ -12,17 +14,25 @@ class Listener {
 }
 
 export type KuzzleSDKEvents =
+  | "connect"
+  | "reconnect"
   | "connected"
   | "reconnected"
+  | "reAuthenticated"
   | "tokenExpired"
   | "loginAttempt"
   | "discarded"
+  | "disconnect"
   | "disconnected"
   | "networkError"
+  | "queryError"
+  | "reconnectionError"
+  | "callbackError"
   | "offlineQueuePop"
   | "offlineQueuePush"
-  | "queryError"
-  | "callbackError";
+  | "logoutAttempt"
+  | "websocketRenewalStart"
+  | "websocketRenewalDone";
 /**
  * @todo proper TS conversion
  */
@@ -33,7 +43,7 @@ export class KuzzleEventEmitter {
     this._events = new Map();
   }
 
-  private _exists(listeners, fn) {
+  private _exists(listeners: Listener[], fn: ListenerFunction) {
     return Boolean(listeners.find((listener) => listener.fn === fn));
   }
 
@@ -45,11 +55,16 @@ export class KuzzleEventEmitter {
     return this._events.get(eventName).map((listener) => listener.fn);
   }
 
-  addListener(eventName: KuzzleSDKEvents, listener, once = false) {
+  addListener(
+    eventName: KuzzleSDKEvents,
+    listener: ListenerFunction,
+    once = false
+  ) {
     if (!eventName || !listener) {
       return this;
     }
 
+    // TODO: this check could be safely, when TypeScript type will be completed.
     const listenerType = typeof listener;
 
     if (listenerType !== "function") {
@@ -70,8 +85,18 @@ export class KuzzleEventEmitter {
   }
 
   on(
-    eventName: "connected" | "reconnected" | "tokenExpired",
+    eventName:
+      | "connected"
+      | "reconnected"
+      | "tokenExpired"
+      | "reAuthenticated"
+      | "websocketRenewalStart"
+      | "websocketRenewalDone",
     listener: () => void
+  ): this;
+  on(
+    eventName: "logoutAttempt",
+    listener: (status: { success: true }) => void
   ): this;
   on(
     eventName: "loginAttempt",
@@ -79,21 +104,12 @@ export class KuzzleEventEmitter {
   ): this;
   on(eventName: "discarded", listener: (request: RequestPayload) => void): this;
   on(
-    eventName: "disconnected",
-    listener: (context: {
-      origin:
-        | "websocket/auth-renewal"
-        | "user/connection-closed"
-        | "network/error";
-    }) => void
+    eventName: "disconnect" | "disconnected",
+    listener: (context: { origin: DisconnectionOrigin }) => void
   ): this;
   on(
-    eventName: "networkError",
-    listener: (error: {
-      message: string;
-      status: number;
-      stack?: string;
-    }) => void
+    eventName: "networkError" | "reconnectionError",
+    listener: (error: Error) => void
   ): this;
   on(
     eventName: "offlineQueuePop",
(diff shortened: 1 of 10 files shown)


src/core/KuzzleEventEmitter.ts
```diff
@@ -1,3 +1,6 @@
+import { KuzzleError } from "src/KuzzleError";
+import { Notification, RequestPayload } from "src/types";
```
Aschen, line 2: You need to use relative path otherwise the build will fail later

src/types/Notification.ts
```diff
@@ -62,11 +56,13 @@ export interface BaseNotification extends Notification {
  * Notification triggered by a document change.
  * (create, update, delete)
  */
-export interface DocumentNotification extends BaseNotification {
+export interface DocumentNotification<
+  DocContent extends KDocumentContentGeneric = KDocumentContentGeneric
```
Aschen, line 60: We usually use a capital T to distinguish template types, can you rename it to `TDocContent` or `TContent` please?

Aschen, line 60: We usually use a capital T to distinguish template types, can you rename it to `TDocContent` or `TContent` please?

Aschen (changes requested):
Thanks for your work, specially on the event emitter, it's really appreciated ;-)

Aschen (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-sdk",
-  "version": "7.10.5",
+  "version": "7.10.6",
```
alexandrebouthinon, line 3: This will be updated by our internal release process (same on package-lock updates) 😉 
```suggestion
  "version": "7.10.5",
```

alexandrebouthinon (commented):
Once unit tests fixed and package* files changes reverted. LGTM 👍

package.json
ScreamZ: Haven't changed, it's aschen commit, i'll try to squash

ScreamZ: @alexandrebouthinon @Aschen I was able to fix the "commit" from aschen by switching target to `master` then back to `7-dev`.

FYI : 
- https://stackoverflow.com/a/26986320/5320409
- https://stackoverflow.com/a/45286781/5320409

ScreamZ:
@alexandrebouthinon @Aschen Ok boys I made a lot of changes I've updated my initial post with further context about some changes too. Feel free to ping me

src/core/KuzzleEventEmitter.ts
```diff
@@ -1,8 +1,10 @@
+import { DisconnectionOrigin } from "src/protocols/DisconnectionOrigin";
```
Aschen, line 1: This import should be relative otherwise the build will fail

Aschen (changes requested):

src/core/KuzzleEventEmitter.ts
ScreamZ: Damn, I'm getting caught every time by auto-import.

Aschen (approved):

ScreamZ:
@Aschen @alexandrebouthinon I would like to get some help here because in my workspace tests are not failing.
[asciicast](https://asciinema.org/a/MafkkMI0xWV1ndckC5dfEx2i8)

src/core/KuzzleEventEmitter.ts
```diff
@@ -54,7 +84,50 @@ export class KuzzleEventEmitter {
...
+    eventName: "logoutAttempt",
+    listener: (status: { success: true }) => void
+  ): this;
+  on(
+    eventName: "loginAttempt",
+    listener: (data: { success: boolean; error: string }) => void
+  ): this;
+  on(eventName: "discarded", listener: (request: RequestPayload) => void): this;
+  on(
+    eventName: "disconnect" | "disconnected",
```
Shiranuit, line 107: The `disconnect` event is only emitted by the protocol not the SDK directly, the SDK only emit `disconnected`

test/kuzzle-sdk-test.ts
```diff
@@ -0,0 +1,48 @@
...
+    _id,
+    body,
+    collection,
+    index,
+    jwt,
+    requestId,
+    volatile,
+  }) => {}
+);
+kuzzle.on("disconnect", ({ origin }) => {});
```
Shiranuit, line 33: This is going to throw since the SDK does not know the `disconnect` event, only `disconnected`

Shiranuit (changes requested):
@ScreamZ 
> I would like to get some help here because in my workspace tests are not failing.

When you are running the test in local without building the SDK it seems that it does not run your new test `test/kuzzle-sdk-test.ts`, but if you build the SDK with `npm run build` before running `npm run test` then the test `test/kuzzle-sdk-test.js` is executed.
Which is exactly what happens in the CI, the CI is building the SDK before running the tests, this is where the difference occurs between you and the CI when running the tests.

src/core/KuzzleEventEmitter.ts
ScreamZ: @Shiranuit Alright thanks. 

In fact this `test/kuzzle-sdk-test.ts`  is not meant to be compiled, but let's do this.
About events, we've got a problem which is I need to type all events because events are triggered by the same event listener and same `on` method, even if they are considered private otherwise the code is inconsistent.

I can only suggest two changes : 
1. Allowing listening on private events and marking them as internal and not stable.
2. Expose a new function `onEvent` which is restricted to "public" events. And mark the original `on` as private.

Maybe there is another way to achieve this, but I'll need your insight.
@Shiranuit @Aschen

ScreamZ: @Shiranuit It looks like I've been able to play with step 1

PR #726, opened 2023-04-12
Add index and collection fields to KHit

fmauNeko:
## What does this PR do ?
It adds `index` and `collection` to `KHit<>`.
Those fields are necessary for the multi-search results.

### DIFF
--- a/src/types/KDocument.ts
+++ b/src/types/KDocument.ts
@@ -70,11 +70,13 @@ export interface KHit<TKDocumentContent extends KDocumentContent>
 
   /**
    * Document index
+   * Present only in the case of a multi search
    */
-  index: string;
+  index?: string;
 
   /**
    * Document collection
+   * Present only in the case of a multi search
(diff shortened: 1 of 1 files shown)


src/types/KDocument.ts
```diff
@@ -67,4 +67,14 @@ export interface KHit<TKDocumentContent extends KDocumentContent>
...
+
+  /**
+   * Document index
+   */
+  index: string;
+
+  /**
+   * Document collection
+   */
+  collection: string;
```
Aschen, line 79: Both are not always present so the type should reflect that

Aschen (changes requested):

src/types/KDocument.ts
fmauNeko: Done !

Aschen (approved):

OlivierCavadenti (approved):

PR #731, opened 2023-11-06
Fix empty raw responses parsed as json

fmauNeko:
## What does this PR do ?

Prevent passing empty strings into JSON.parse when using raw responses over HTTP.

### DIFF
--- a/.github/actions/es-lint/action.yml
+++ b/.github/actions/es-lint/action.yml
@@ -4,7 +4,7 @@ runs:
   using: "composite"
   steps:
     - name: Install deps
-      run: npm install
+      run: npm ci
       shell: bash
     - name: Run lint
       run: npm run test:lint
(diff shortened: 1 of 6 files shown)


src/protocols/Http.ts
```diff
@@ -405,6 +405,10 @@ export default class HttpProtocol extends KuzzleAbstractProtocol {
             );
           }
 
+          if (response.body === "") {
+            return "";
+          }
```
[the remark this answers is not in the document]
fmauNeko: That would be the ideal way to do, but that requires a change on the backend side.
Indeed, the backend still transmits a `Content-Type: application/json` even with an empty response.
That would require a more involved change in Kuzzle.

rolljee (approved):

.github/actions/es-lint/action.yml
```diff
@@ -4,7 +4,7 @@ runs:
   using: "composite"
   steps:
     - name: Install deps
-      run: npm i -g npm && npm install
+      run: npm install
```
rolljee, line 7: can we use npm ci here ?

fmauNeko: Probably

PR #734, opened 2023-11-30
fix: (ProfilePolicy.ts) restrictedTo was mistyped (#733)

Juiced66:
## What does this PR do ?

Fix ProfilePolicy type (restrictedTo is an array of restrictions)


rolljee:
This code needs to go to beta first

PR #739, opened 2024-01-19
Update Notification interface to match correct API

ScreamZ:
Improve document payload according to real documentation

https://docs.kuzzle.io/core/2/api/payloads/notifications/#document-notification

Can you ping me on release @rolljee  @etrousset ? Thanks :)

### DIFF
--- a/src/types/Notification.ts
+++ b/src/types/Notification.ts
@@ -1,5 +1,5 @@
 import { JSONObject } from "./JSONObject";
-import { KDocument, KDocumentContentGeneric } from ".";
+import { KDocumentContentGeneric } from ".";
 
 /**
  * Enum for notification types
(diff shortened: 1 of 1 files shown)


src/types/Notification.ts
```diff
@@ -62,7 +62,21 @@ export interface DocumentNotification<
...
+     */
+    _source: TDocContent;
+    /**
+     * Document unique ID.
+     * `null` if the notification is from a real-time message.
+     */
+    _id: string | null;
+    /**
+     * List of fields that have been updated (only available on document partial updates).
+     */
```
ScreamZ, line 77: Not sure what « only available on document partial updates » means. Is it missing? Is it `null`?

ScreamZ: I just tested, and the field is missing, so the type is good.

etrousset (approved):

PR #743, opened 2024-04-12
fix: adding id checking to replace old jwt in queue

thomas-mauran:
## What does this PR do ?
Fix the queue jwt token when it contains old ones

  - Step 1 :
  - Step 2 :
  - Step 3 :


thomas-mauran:
I think this can be closed cause solved in #744

PR #744, opened 2024-05-17
Reapply current JWT before executing offlineQueue requests

alexandrebouthinon:
## What does this PR do ?
Reapply current JWT before executing offlineQueue requests.

Since the final user could use the authenticator with a short JWT TTL or being disconnected for a big amount of time from his Kuzzle backend.

### DIFF
--- a/.github/workflows/pull_request.workflow.yml
+++ b/.github/workflows/pull_request.workflow.yml
@@ -85,9 +85,6 @@ jobs:
      name: Documentation Snippet Tests
      needs: [unit-tests]
      runs-on: ubuntu-20.04
-     strategy:
-      matrix:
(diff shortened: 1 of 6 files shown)


.github/workflows/pull_request.workflow.yml
```diff
@@ -85,9 +85,6 @@ jobs:
      name: Documentation Snippet Tests
      needs: [unit-tests]
      runs-on: ubuntu-20.04
-     strategy:
-      matrix:
-        node-version: ["16", "18"]
```
alexandrebouthinon, line 90: This is useless since the version used to test the snippets is the version set in the docker compose file above

fmauNeko (approved):

PR #745, opened 2024-09-17
feat: support trigger events option

Juiced66:
This PR is intended to support triggerEvents request parameter for embeded sdk. 

It is unit tested and will be functionnal tested when released on kuzzle side. 

A lot of doc as been realigned with code

JIRA: [KZLPRD-579](https://kuzzle.atlassian.net/browse/KZLPRD-579) 

[KZLPRD-579]: https://kuzzle.atlassian.net/browse/KZLPRD-579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/controllers/Bulk.ts
+++ b/src/controllers/Bulk.ts
@@ -31,7 +31,7 @@ export class BulkController extends BaseController {
     query: JSONObject = {},
     options: ArgsBulkControllerDeleteByQuery = {}
   ): Promise<number> {
-    const request: Omit<RequestPayload, "controller"> = {
+    const request = {
       action: "deleteByQuery",
       body: query,
       collection,
@@ -108,7 +108,7 @@ export class BulkController extends BaseController {
       };
     }>;
   }> {
-    const request: Omit<RequestPayload, "controller"> = {
+    const request = {
       action: "import",
       body: { bulkData },
       collection,
@@ -141,7 +141,7 @@ export class BulkController extends BaseController {
     changes: JSONObject,
     options: ArgsBulkControllerUpdateByQuery = {}
   ): Promise<number> {
-    const request: Omit<RequestPayload, "controller"> = {
+    const request = {
       action: "updateByQuery",
       body: { changes, query },
       collection,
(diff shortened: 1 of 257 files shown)


etrousset (approved):

rolljee (commented):
Everything look good and updated.

There is just something odd about the :any you added to the request.

Since you have the type for the KuzzleRequest isn't that possible to put the type instead of any ?

Juiced66:
> Why didn't you put this (duplicated) piece of code in the `query` method directly?
> 
> ```js
> if (options.triggerEvents !== undefined) {
>   request.triggerEvents = options.triggerEvents;
> }
> ```
> 
> I see that stuff like this is already done in this method:
> 
> https://github.com/kuzzleio/sdk-javascript/blob/83e434b679f884604e567976967f31562ec5ba32/src/Kuzzle.ts#L901-L906

I've thinked about that, it was an option but not all methods got options and it could lead to useless code for those. I thought it was less readable to apply it were needed but more optimized. If we decide to apply it directly on query it is doable for sure

Juiced66:
> Everything look good and updated.
> 
> There is just something odd about the :any you added to the request. Since you have the type for the KuzzleRequest isn't that possible to put the type instead of any ?
> 
> Or like Alex said, it could be possible to avoid duplication at the Kuzzle.ts file level

I could use `Omit<RequestPayload, "controller">`  wiches weird too but maybe more acceptable. Controller is abstracted and taken from controller name and it leads to ts errors without omitting

src/controllers/Bulk.ts
```diff
@@ -62,62 +66,33 @@ export class BulkController extends BaseController {
     bulkData: Array<JSONObject>,
     options: ArgsBulkControllerImport = {}
   ): Promise<{
-    /**
-     * Array of successfully executed actions
-     */
     successes: Array<{
```
sebtiz13, line 69: Why remove this doc comments ?

sebtiz13 (approved):

Juiced66:
For maintainability purpose and after doing pro and cons, I'll stick to maintainability purpose where the query solution is better than little optim nit picking

rolljee:
LIttle UP as this need to be merged 💪🏼

Juiced66:
will close this to open a Doc only PR

PR #753, opened 2024-10-16
feat: add options parameter to controller functions containing a query

cyrng:
## What does this PR do ?

This PR add missing options parameter to controller methods that include a query so the option allows to use trigger event.

  - Step 1 :
  - Step 2 :
  - Step 3 :


src/controllers/Auth.ts
```diff
@@ -505,7 +510,7 @@ export class AuthController extends BaseController {
         {
           action: "logout",
           cookieAuth: this.kuzzle.cookieAuthentication,
-          ...options
+          ...options,
```
etrousset, line 513: Are you sure you expands the options in the right object?

rolljee (approved):

rolljee (approved):

PR #756, opened 2024-11-20
fix: Websocket propper error when the payload size is too big

thomas-mauran:
## What does this PR do ?

If the sdk is using a WebSocket and trying to send a payload > to the maxPayload defined in the backend ( by default 1mb but can be extended in the kuzzlerc) we are getting a Network error with no real indication on what happened instead of a proper 413 error max payload size exceeded. 

This bug is happening because of the way the uWebSocket library is handling the maxPayload case. If a payload is received in the socket with a size over the maxPayload the connection gets automatically closed [see here](https://github.com/uNetworking/uWebSockets.js/blob/b3948e68b7625b5f94f161aa563500cc73f4aef2/docs/index.d.ts#L229). The connection being closed doesn't allow us to provide a propper error message on the backend side. Instead we have to adapt the sdk by doing the following: 

When we open a connection we also send a request to the backend `controller: server; action: getConfig` where reside the defined maxRequestSize of the server. We store this value in this.maxPayloadSize in the WebSockerProtocol Class. Each time we perform a request, we have a if condition checking if the payload we are trying to send is not over this value. If so, we properly close the connection before sending the message as it should be done.

here is the new type of error message we get now (here the server limit is 2mb): 

```bash
 
> test@1.0.0 start
> node main.js

Network Error:  {
  error: Error: Payload size exceeded the maximum allowed by the server 2097152 bytes
      at WebSocketProtocol.send (/home/thomas/Documents/sdk-javascript/src/protocols/WebSocket.js:183:27)
      at WebSocketProtocol.query (/home/thomas/Documents/sdk-javascript/src/protocols/abstract/Base.js:137:14)
      at Proxy._timeoutRequest (/home/thomas/Documents/sdk-javascript/src/Kuzzle.js:773:34)
      at Proxy.query (/home/thomas/Documents/sdk-javascript/src/Kuzzle.js:599:21)
      at run (file:///home/thomas/Documents/playground/sdk-test/main.js:24:35)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
}
Network error: request was sent but no response has been received
```

  - Step 1 : using kourou scaffold a new project
  - Step 2 : create the following script
  ```js
  import kuzzleSDK from "kuzzle-sdk";
import fs from "fs";

const { Kuzzle, WebSocket } = kuzzleSDK;

// Replace 'localhost' with your Kuzzle server hostname if needed
const kuzzle = new Kuzzle(new WebSocket("localhost"));

kuzzle.on("networkError", (error) => {
  console.error("Network Error: ", error);
});

const run = async () => {
  try {
    await kuzzle.connect();
    // console.log("Connected to Kuzzle server");

    const data = fs.readFileSync("./working.txt");
    // const data = fs.readFileSync("./big-file.png");
    // console.log("here", data);

    const response = await kuzzle.query({
      controller: "test",
      action: "sendFile",
      body: {
        body: {
          file: data,
        },
      },
    });
    // console.log("File sent", response);
  } catch (error) {
    console.error(error.message);
  } finally {
    kuzzle.disconnect();
  }
};

run();

```

You might also have a package.json like the following 
```json
{
  "name": "test",
  "version": "1.0.0",
  "main": "main.js",
  "type": "module",
  "scripts": {
    "start": "node main.js"
  },
  "dependencies": {
    "kuzzle-sdk": "file:../../sdk-javascript"
  }
}
```
  - Step 3 : clone the sdk
  - Step 4 : npm i && npm run build it
  - Step 5 : npm link it into your scaffolded kuzzle where the script reside
  - Step 6 : npm i and npm run start (make sure your scaffold kuzzle is running)
  
  This command should try to send a request to the backend using websocket. In the code you must also add a file under 1mb and one > to 1mb, big-file.png is an example where I used an heavy image to reproduce the bug.

### DIFF
--- a/src/protocols/WebSocket.ts
+++ b/src/protocols/WebSocket.ts
@@ -243,9 +243,11 @@ export default class WebSocketProtocol extends BaseProtocolRealtime {
    * @param {Object} payload
    */
   send(request: RequestPayload, options: JSONObject = {}) {
-    if (this.maxPayloadSize !== null && Buffer.byteLength(JSON.stringify(request), 'utf8') > this.maxPayloadSize) {
-      
-      const error: any= new Error(
+    if (
+      this.maxPayloadSize !== null &&
+      Buffer.byteLength(JSON.stringify(request), "utf8") > this.maxPayloadSize
+    ) {
+      const error: any = new Error(
         `Payload size exceeded the maximum allowed by the server ${this.maxPayloadSize} bytes`
       );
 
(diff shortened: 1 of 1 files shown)


QuentinRousselet (approved):

inkedsquid:
Après discussion avec Eric et Ricky, on a décidé de simplifier le traitement et de seulement log l'erreur. Pour le côté client, on changera le message d'erreur pour qu'il soit plus compréhensible

